PR #3, opened 2020-01-06
Implement Solidity encoding and decoding 

danielakhterov:


contract_function_selector.go
```diff
@@ -0,0 +1,224 @@
...
+	aInt256   argument = "int256"
+	aUint32   argument = "uint32"
+	aUint64   argument = "uint64"
+	aUint256  argument = "uint256"
+	aBytes    argument = "bytes"
+	aFunction argument = "function"
+	aAddress  argument = "address"
+)
+
+func NewContractFunctionSelector(function *string) ContractFunctionSelector {
```
mehcode: It might be better to have `""` mean `nil` for the alg here

--

As you can't just do `NewContractFunctionSelector(&"name")` in go ( need temp var )

PR #4, opened 2020-01-06
Keygen and Crypto Parity

QuestofIranon:
This adds key generation using SLIP-10/BIP-32 Child Key derivation and mnemonics and brings existing pieces up to par with other SDKs. #9 will focus on adding in more new features.


QuestofIranon:
Setting as a Work In Progress, needs a public derive function on the keys.

crypto.go
```diff
@@ -99,6 +130,48 @@ func Ed25519PublicKeyFromString(s string) (Ed25519PublicKey, error) {
...
+	return Ed25519PublicKey{
+		keyData: bytes,
+	}, nil
+}
+
+func ed25519PublicKeyFromProto(proto proto.Key) (Ed25519PublicKey, error) {
+	rawKey := proto.GetEd25519()
+
+	if rawKey == nil {
+		return Ed25519PublicKey{}, fmt.Errorf("not an ed25519 key")
```
danielakhterov: Perhaps a more in depth error message?

danielakhterov:
Could we also get `publicKeyFromProto`? This is different than the `ed25519FromProto()` https://github.com/hashgraph/hedera-sdk-js/blob/3276de15e7e2cd0b7380050141183ec5a0d45656/src/crypto/PublicKey.ts#L7

QuestofIranon:
@danielakhterov 

> Could we also get `publicKeyFromProto`? This is different than the `ed25519FromProto()` https://github.com/hashgraph/hedera-sdk-js/blob/3276de15e7e2cd0b7380050141183ec5a0d45656/src/crypto/PublicKey.ts#L7

As we discuessed, this can be done in a future pull request (it's beyond the scope of this one)

PR #11, opened 2020-01-10
KeyStores, KeyLists, and other Crypto Parity issues

### LINKED ISSUE #14: Add tests to ensure mnemonics work correctly
> [...] unable to get the mnemonic data (looks like its not passed from the ed25519 library) [...]

was reported on Discord

QuestofIranon:
closes #14, #13

### DIFF
--- a/crypto.go
+++ b/crypto.go
@@ -9,6 +9,8 @@ import (
 	"fmt"
 	"golang.org/x/crypto/ed25519"
 	"golang.org/x/crypto/pbkdf2"
+	"io"
+	"io/ioutil"
 	"strings"
 
 	"github.com/hashgraph/hedera-sdk-go/proto"
@@ -108,6 +110,19 @@ func Ed25519PrivateKeyFromString(s string) (Ed25519PrivateKey, error) {
 	return Ed25519PrivateKey{}, fmt.Errorf("invalid private key with length %v", len(s))
 }
 
+func Ed25519PrivateKeyFromKeystore(ks []byte, passphrase string) (Ed25519PrivateKey, error) {
+	return parseKeystore(ks, passphrase)
+}
+
+func Ed25519PrivateKeyReadKeystore(source io.Reader, passphrase string) (Ed25519PrivateKey, error) {
+	keystoreBytes, err := ioutil.ReadAll(source)
+	if err != nil {
+		return Ed25519PrivateKey{}, fmt.Errorf("failed to read keystore from provided source")
+	}
+
+	return Ed25519PrivateKeyFromKeystore(keystoreBytes, passphrase)
+}
+
 func Ed25519PublicKeyFromString(s string) (Ed25519PublicKey, error) {
 	switch len(s) {
 	case 64: // raw public key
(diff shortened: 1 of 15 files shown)


crypto.go
```diff
@@ -108,6 +110,19 @@ func Ed25519PrivateKeyFromString(s string) (Ed25519PrivateKey, error) {
...
 }
 
+func Ed25519PrivateKeyFromKeystore(ks []byte, passphrase string) (Ed25519PrivateKey, error) {
+	return parseKeystore(ks, passphrase)
+}
+
+func Ed25519PrivateKeyReadKeystore(source io.Reader, passphrase string) (Ed25519PrivateKey, error) {
+	keystoreBytes, err := ioutil.ReadAll(source)
+	if err != nil {
+		return Ed25519PrivateKey{}, fmt.Errorf("failed to read keystore from provided source")
```
mehcode: Return `err` directly

crypto.go
```diff
@@ -108,6 +110,19 @@ func Ed25519PrivateKeyFromString(s string) (Ed25519PrivateKey, error) {
 	return Ed25519PrivateKey{}, fmt.Errorf("invalid private key with length %v", len(s))
 }
 
+func Ed25519PrivateKeyFromKeystore(ks []byte, passphrase string) (Ed25519PrivateKey, error) {
+	return parseKeystore(ks, passphrase)
+}
+
+func Ed25519PrivateKeyReadKeystore(source io.Reader, passphrase string) (Ed25519PrivateKey, error) {
```
mehcode: This should probably just be `ReadKeystore`. We can extend it to support other types in the future. This would mirror to a generic `PrivateKey.readKeystore` in say Java.

crypto.go
```diff
@@ -188,6 +203,25 @@ func (sk Ed25519PrivateKey) Bytes() []byte {
...
+func (sk Ed25519PrivateKey) WriteKeystore(destination io.Writer, passphrase string) error {
+	keystore, err := sk.Keystore(passphrase)
+	if err != nil {
+		return err
+	}
+
+	_, err = destination.Write(keystore)
+
+	if err != nil {
+		return fmt.Errorf("could not write keystore to destination")
```
mehcode: Just return the `err`

keystore.go
```diff
@@ -0,0 +1,201 @@
...
+	"crypto/sha256"
+	"crypto/sha512"
+	"encoding/hex"
+	"encoding/json"
+	"fmt"
+	"golang.org/x/crypto/pbkdf2"
+	"io"
+)
+
+type Keystore struct {
```
mehcode: `Keystore` should not be a public type

keystore.go
```diff
@@ -0,0 +1,201 @@
...
+		return nil, err
+	}
+
+	return b, nil
+}
+
+func newKeystore(privateKey []byte, passphrase string) ([]byte, error) {
+	salt, err := randomBytes(saltLen)
+	if err != nil {
+		return nil, fmt.Errorf("could not generate salt bytes")
```
mehcode: Return `err`

crypto.go
QuestofIranon: @mehcode There isn't a very clean (from the perspective of the sdk user) way to make it generic, my thought was that the inner keystore function would eventually be made more generic and each key type will have the ReadKeystore function.

My concern with that is multi key keystores though -- not sure if we will need those.

PR #36, opened 2020-02-06
Pem support for ed25519 private keys

QuestofIranon:


mehcode:
Merging, we will work on the CI failures today

PR #37, opened 2020-02-11
Split ErrHederaStatus into ErrHederaPrecheckStatus, ErrHederaReceiptStatus, and ErrHederaRecordStatus

QuestofIranon:


errors.go
```diff
@@ -54,23 +58,48 @@ func (e ErrHederaNetwork) Error() string {
 	return fmt.Sprintf("transport error occurred while accessing the Hedera network: %s", e.Error())
 }
 
-type ErrHederaStatus struct {
```
danielakhterov: I thought we're not deleting the old type. Instead I thought the new error types would extend this error type. IIRC, it would be something like this. This should allow you access to the `Status` underneath.
```go
type ErrHederaPrecheckStatus struct {
    ErrHederaStatus
}
```

errors.go
```diff
@@ -54,23 +58,48 @@ func (e ErrHederaNetwork) Error() string {
...
+// ErrHederaPrecheckStatus is returned by Transaction.Execute and QueryBuilder.Execute if an exceptional status is
+// returned during network side validation of the sent transaction.
+type ErrHederaPrecheckStatus struct {
```
danielakhterov: This should also include `TransactionId TransactionID`?

errors.go
QuestofIranon: This can't be done properly in Go and also isn't needed since Go isn't on 1.0.0 yet.

PR #44, opened 2020-02-15
GetCost() on transaction

QuestofIranon:
may need to update protos


transaction_builder.go
```diff
@@ -3,22 +3,26 @@ package hedera
...
 	"github.com/hashgraph/hedera-sdk-go/proto"
+	"math"
 	"time"
 )
 
 type TransactionBuilder struct {
 	pb *proto.TransactionBody
+
+	// unfortunately; this is required
+	setTXFee bool
```
QuestofIranon: I really don't like this, but go defaults ints to 0 and I couldn't think of a good way to determine if the 0 was set purposely or by the user/api. Otherwise, unless the client has the `maxTransactionFee` set to 0 it will always pull that if the transactionID has not been set.

This might cause issues with marshalling/unmarshalling though so I'm very open to better ways to handle it.

QuestofIranon: I'm going to keep this flag and make it work more like how the [JS SDK handles it](https://github.com/hashgraph/hedera-sdk-js/blob/06057d1cf07941123ca52e79877f3479f7505eb6/src/TransactionBuilder.ts#L97)

PR #66, opened 2020-04-02
Downgrade the protobuf package

QuestofIranon:
Moves the protobuf package off of a release candidate and back to a stable version.


mehcode:
What. Why are the tests passing now? Was it the protobuf rc that was breaking things in a random place?

Issue #84, opened 2020-08-14
Appending transaction body signatures to a transaction


failfmi:
Is there currently a way in the SDK to append transaction body signatures to a transaction?
Scenario: send unsigned transaction & transaction body to a party, which will sign it and later append the signature to the transaction, so that it can be broadcasted?

mehcode:
Please look to this example: https://github.com/hashgraph/hedera-sdk-go/blob/master/examples/multi_app_transfer/main.go

---

You can marshall the transaction to bytes, transfer through the network, unmarshall to a Transaction, and sign and execute.

failfmi:
Problem is party signer does not know anything related to Hedera, calls `ed25519.Sign(privateKey, message)`.

PR #87, opened 2020-08-21
Feature(transaction): Get Signature Pairs, Get Body and Body Bytes

### LINKED ISSUE #85: Accessing Transaction Body
Is there a way in the SDK to get the transaction body, so that I can get its transaction data?
I am asking that with the idea of rebuilding an object with which I have made the actual transaction in first place.

Example:
```JSON
{"type": "CRYPTOTRANSFER", "account": {"address": "0.0.123352"}, "amount": "-100"},
{"type": "CRYPTOTRANSFER", "account": {"address": "0.0.123518"}, "amount": "+100"}
```
to
```Go
hedera.NewCryptoTransferTransaction()
.AddSender(sender, senderAmount)
.AddRecipient(recipient, recipientAmount)
```

failfmi:
**Detailed description**:
- Add tx `SignaturePairs` method
- Add tx `BodyBytes` method
- Update tx `Body` method to be accessible outside package

**Which issue(s) this PR fixes**:
Fixes #85

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -156,3 +156,18 @@ func newErrPingStatus(e error) ErrPingStatus {
 func (e ErrPingStatus) Error() string {
 	return fmt.Sprintf("error occured during ping")
 }
+
+// ErrSignatureVerification is returned by Transaction.AppendSignature(Ed25519PublicKey, []byte)
+// if signature verification fails.
+type ErrSignatureVerification struct {
+	message string
+}
+
+func newErrSignatureVerification(format string, a ...interface{}) ErrSignatureVerification {
+	return ErrSignatureVerification{fmt.Sprintf(format, a...)}
+}
+
+// Error() implements the Error interface
+func (e ErrSignatureVerification) Error() string {
+	return e.message
+}
--- a/errors.go
+++ b/errors.go
@@ -156,18 +156,3 @@ func newErrPingStatus(e error) ErrPingStatus {
 func (e ErrPingStatus) Error() string {
 	return fmt.Sprintf("error occured during ping")
 }
-
-// ErrSignatureVerification is returned by Transaction.AppendSignature(Ed25519PublicKey, []byte)
-// if signature verification fails.
-type ErrSignatureVerification struct {
-	message string
-}
-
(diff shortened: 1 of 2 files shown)


failfmi:
As I have said in #84 , the party signer does not know anything related to Hedera, and the faced scenario is that party signer signs the `tx body` and sends it to another service, which appends the signature to the `Transaction`.

`BodyBytes` is needed so that transaction body is extracted from the `Transaction`, allowing to be externally signed.

`Body` and `SignaturePairs` are related to #85 , which I need in order to rebuild a specific JSON from which the `Transaction` has been built in first place (getting all transfers - `accountAmounts`, and the tx signers if any).

Daniel-K-Ivanov:
@mehcode what is the status of this PR and are you planning to expose the `BodyBytes` in the new version of the SDK?

mehcode:
> @mehcode what is the status of this PR and are you planning to expose the BodyBytes in the new version of the SDK?
At the moment we are using a forked version of the SDK as we do not have any other way to externally sign the transactions.


`SignWith` is exactly the method you are looking for. It exists on master of the SDK.

```go
tx.SignWith(myPublicKey, func(bodyBytes []byte) []byte {
  // take bodyBytes and do whatever you want to produce your own signature
})
```

Daniel-K-Ivanov:
> > @mehcode what is the status of this PR and are you planning to expose the BodyBytes in the new version of the SDK?
> > At the moment we are using a forked version of the SDK as we do not have any other way to externally sign the transactions.
> 
> `SignWith` is exactly the method you are looking for. It exists on master of the SDK.
> 
> ```go
> tx.SignWith(myPublicKey, func(bodyBytes []byte) []byte {
>   // take bodyBytes and do whatever you want to produce your own signature
> })
> ```

Yes, after your comment in august we started using `SignWith`, however, we still need `BodyBytes` to be public. You can see the files changes in this PR. They are only for making some private functions public in order to accommodate `external` signing of transactions. In other terms, we have a use case in which the signer of the transactions has only a PK and requires the bytes to sign. We need to construct the transaction and pass only the bytes to the signer without him ever needing to use the SDK.

PR #99, opened 2020-11-19
Update consensus_pub_sub example

gregscullard:
Add a 10s delay prior to mirror subscription to enable topic propagation to mirrors.


danielakhterov:
This PR is no longer required because transactions produce a  `TransactionResponse` after execution which allows us to reuse the same node ID for subsequent transactions and queries.

PR #114, opened 2021-01-02
Update transaction_id.go to resolve overflow error

jpersofsky:
I got the following error when using gomobile bind:
transaction_id.go:21:42: constant 5000000000 overflows int

I changed the types to int64 and that seems to resolve the issue.


danielakhterov (approved):
LGTM. Thanks for the PR @jpersofsky

PR #143, opened 2021-05-10
fix: `Hbar.As()` should return a `float64` and use division

### LINKED ISSUE #135: Wrong conversion from tinybar to other unit
https://github.com/hashgraph/hedera-sdk-go/blob/914d7893fa531636bc0d64a24566555248989994/hbar.go#L42-L44

It should be division

danielakhterov:
Closes #135

### DIFF
--- a/hbar.go
+++ b/hbar.go
@@ -39,8 +39,8 @@ func (hbar Hbar) AsTinybar() int64 {
 	return hbar.tinybar
 }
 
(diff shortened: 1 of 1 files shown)


danielakhterov:
Merging because the reason PR failed is due to transactions/queries in tests not having `SetNodeAccountIDs(env.NodeAccountIDs)`, not because of the change this PR made.

PR #163, opened 2021-06-02
make ed25519 key prefixes public

safinbot:


crypto.go
```diff
@@ -20,8 +20,8 @@ import (
 	"github.com/youmark/pkcs8"
 )
 
-const ed25519PrivateKeyPrefix = "302e020100300506032b657004220420"
-const ed25519PubKeyPrefix = "302a300506032b6570032100"
+const Ed25519PrivateKeyPrefix = "302e020100300506032b657004220420"
```
danielakhterov, line 23: Since you're making these public, can you name these `PrivateKeyDerPrefix` and `PublicKeyDerPrefix`. This will allow us to more easily port this in the Java and JS SDK's.

danielakhterov (changes requested):

mehcode:
I don't understand why you need this. Can you help me understand your use case?

PR #166, opened 2021-06-09
refactor: `TopicMessageQuery`

### LINKED ISSUE #132: `TransactionReceipt` should error when status is not `SUCCESS` and return a `TransactionReceipt` *within* the error
When submitting a schedule sign transaction, which will not update the signatures of the scheduled transaction, the transaction response `GetReceipt` returns an error `exceptional receipt status NO_NEW_VALID_SIGNATURES`

1. The receipt request itself should not resolve with an error, but more like with a transaction receipt, which has the status

### LINKED ISSUE #169: `TopicMessageQuery` refactor

danielakhterov:
Resolves #132 
Resolves #169

### DIFF
--- a/topic_message_query.go
+++ b/topic_message_query.go
@@ -12,15 +12,23 @@ import (
 )
 
 type TopicMessageQuery struct {
-	pb           *mirror.ConsensusTopicQuery
-	errorHandler func(stat status.Status)
+	pb                *mirror.ConsensusTopicQuery
+	errorHandler      func(stat status.Status)
+	completionHandler func()
+	retryHandler      func(err error) bool
+	counter           uint64
+	attempt           uint64
+	maxAttempts       uint64
+	limit             *uint64
 }
 
 func NewTopicMessageQuery() *TopicMessageQuery {
-	pb := mirror.ConsensusTopicQuery{}
 	return &TopicMessageQuery{
-		pb:           &pb,
-		errorHandler: nil,
+		pb:                &mirror.ConsensusTopicQuery{},
+		maxAttempts:       maxAttempts,
+		errorHandler:      nil,
+		retryHandler:      defaultRetryHandler,
+		completionHandler: nil,
 	}
 }
 
@@ -64,12 +72,16 @@ func (query *TopicMessageQuery) GetEndTime() time.Time {
 }
 
 func (query *TopicMessageQuery) SetLimit(limit uint64) *TopicMessageQuery {
-	query.pb.Limit = limit
+	query.limit = &limit
 	return query
 }
 
 func (query *TopicMessageQuery) GetLimit() uint64 {
-	return query.pb.Limit
+	if query.limit != nil {
+		return *query.limit
+	} else {
+		return 0
+	}
 }
 
 func (query *TopicMessageQuery) SetErrorHandler(errorHandler func(stat status.Status)) *TopicMessageQuery {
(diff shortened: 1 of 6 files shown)


andrix10 (approved):

steven-sheehy:
A lot of the same problems I noted in the [js](https://github.com/hashgraph/hedera-sdk-js/pull/399) implementation surface here as well.

topic_message_query.go
```diff
@@ -64,79 +72,87 @@ func (query *TopicMessageQuery) GetEndTime() time.Time {
...
+
 				if err != nil {
-					cancel()
-					grpcErr, ok := status.FromError(err)
-					if query.errorHandler == nil && ok {
-						query.errorHandler(*grpcErr)
-					}
+					handle.Unsubscribe()
+					callErrorHandlerWithGrpcStatus(err, query.errorHandler)
+					subClient = nil
```
xin-hedera, line 131: should break out the for loop, otherwise the next line `resp, err := subClient.Rev()` will cause nil pointer derference

also move `query.attemp += 1` to after this if block

topic_message_query.go
```diff
@@ -64,79 +72,87 @@ func (query *TopicMessageQuery) GetEndTime() time.Time {
...
-			cancel()
-			grpcErr, ok := status.FromError(err)
-			if query.errorHandler == nil && ok {
-				query.errorHandler(*grpcErr)
-			}
-		}
 
 		for {
-			if resubscribe {
+			if query.attempt <= query.maxAttempts && subClient == nil {
```
xin-hedera, line 118: with the logic from line 138 - 145, it's possible that

`query.attempt > query.maxAttemps` and `subClient == nil`, the it falls to line 135, nil pointer deference

there the if condition should be just

`if subClient == nil {`

also have to change line 138

topic_message_query.go
```diff
@@ -64,79 +72,87 @@ func (query *TopicMessageQuery) GetEndTime() time.Time {
...
-					}
-					break
-				} else {
-					cancel()
-					grpcErr, ok := status.FromError(err)
-					if query.errorHandler == nil && ok {
-						query.errorHandler(*grpcErr)
-					}
-					break
+				if query.attempt <= query.maxAttempts && query.retryHandler(err) {
```
xin-hedera, line 138: change it to

`if query.attempt < query.maxAttempts && query.retryHandler(err) {`

topic_message_query.go
```diff
@@ -64,79 +72,87 @@ func (query *TopicMessageQuery) GetEndTime() time.Time {
...
+					handle.Unsubscribe()
+					subClient = nil
+
+					delay := 250.0 * math.Pow(2.0, float64(query.attempt))
+					time.Sleep(time.Duration(delay) * time.Millisecond)
+					query.attempt += 1
```
xin-hedera, line 144: should move `query.attempt += 1` to the top block of the for loop

topic_message_query.go
```diff
@@ -64,79 +72,87 @@ func (query *TopicMessageQuery) GetEndTime() time.Time {
...
-				} else {
-					cancel()
-					grpcErr, ok := status.FromError(err)
-					if query.errorHandler == nil && ok {
-						query.errorHandler(*grpcErr)
-					}
-					break
+				if query.attempt <= query.maxAttempts && query.retryHandler(err) {
+					handle.Unsubscribe()
+					subClient = nil
```
xin-hedera, line 140: nit:

these two lines can be moved to the top of `if err != nil` block, no need to duplicate `handle.Unsubscribe()`

Issue #150, opened 2021-05-21
TransferTransaction.GetHbarTransfers should return slice not map


xin-hedera:
https://github.com/hashgraph/hedera-sdk-go/blob/b2f3ced77423d608f26eaddba557abafd2401f10/transfer_transaction.go#L57-L75

- `TransferTransaction.AddHbarTransfer` allows multiple `AccountAmount` from the same account
- `TransferTransaction.GetHbarTransfers` returns a map instead of a slice, does not match the input (number of transfers, order as in protobuf, and etc). In addition, in case of multiple `AccountAmount` from the same account, the implementation of `GetHbarTransfers` just returns the last amount.

danielakhterov:
We decided to make `TransferTransaction.GetHbarTransfers()` return a map since we think it's easier for the user to see the full amount being transferred into/from an account. Forcing a user to search and add the values together _imo_ isn't a great API, although I do agree that the get/set methods being different is an annoyance. Also, _iirc_ when fetching transaction records, Hedera also combines all the transfers for each account into a single number, so this API would be consistent with that.

 As for the SDK incorrectly using the last value instead of adding to the existing value, that is definitely a bug that should be fixed.

xin-hedera:
there are use cases that require such symmetric operation or in other words, reconstruct the transfer list. Specifically,  coinbase rosetta-api validation.

The validation fails when the transfer list built from the constructed `CryptoTransferTransaction` is in the aggregated form, and the original operations (in rosetta term) have multiple transfers from the same account. The order is not important though.

xin-hedera:
I think if we follow the java sdk implementation, i.e., aggregate the hbar transfers in AddHbarTransfer and save the aggregated transfers into the protobuf, it'll solve the problem I hit.

danielakhterov:
I'm a bit confused. In Java SDK we also use a `Map`
```java
    private final Map<AccountId, Hbar> hbarTransfers = new HashMap<>();
```
This would still group the transfers from/to one account into a single field instead of separate.

xin-hedera:
yes, you are right about the java SDK.

What I'm proposing is make the go SDK consistent with the java SDK on when to group the hbar transfers: in java, it's grouped in **Add** v.s. in go it's grouped in **Get**.

This does not require API change and should make the rosetta scenario work.

[...]

PR #170, opened 2021-06-10
fix: `TransactionTransfer.AddHbarTransfer()` 

danielakhterov:
Resolves: #150

### DIFF
--- a/transfer_transaction.go
+++ b/transfer_transaction.go
@@ -8,8 +8,9 @@ import (
 
 type TransferTransaction struct {
 	Transaction
-	pb           *proto.CryptoTransferTransactionBody
-	tokenIndexes map[TokenID]int
+	pb            *proto.CryptoTransferTransactionBody
+	tokenIndexes  map[TokenID]int
+	hbarTransfers map[AccountID]Hbar
 }
 
 func NewTransferTransaction() *TransferTransaction {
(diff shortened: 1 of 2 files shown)


xin-hedera (changes requested):
LGTM

the test failed because when parsing a `TransferTransaction` from marshaled bytes, the `hbarTransfers` map needs to be populated from the protobuf.

perhaps the same applies to `tokenIndexes`

transfer_transaction.go
```diff
@@ -110,6 +150,9 @@ func (transaction *TransferTransaction) Schedule() (*ScheduleCreateTransaction,
 }
 
 func (transaction *TransferTransaction) constructScheduleProtobuf() (*proto.SchedulableTransactionBody, error) {
+	transaction.setHbarTransfersOnProtobuf()
```
andrix10, line 153: transaction.buildHbarTransfers() would be a better name for this in my opinion

transfer_transaction.go
```diff
@@ -110,6 +150,9 @@ func (transaction *TransferTransaction) Schedule() (*ScheduleCreateTransaction,
...
+	transaction.setTokenTransfersOnProtobuf()
```
andrix10, line 154: transaction.buildTokenTransfers() would be a better name for this in my opinion

andrix10 (approved):

PR #173, opened 2021-06-12
fix the `TopicMessageQuery` refactor

danielakhterov:
Resolves comments made on #166

### DIFF
--- a/topic_message_query.go
+++ b/topic_message_query.go
@@ -2,8 +2,8 @@ package hedera
 
 import (
 	"context"
+	"io"
 	"math"
-	"sync"
 	"time"
 
 	"github.com/hashgraph/hedera-sdk-go/v2/proto/mirror"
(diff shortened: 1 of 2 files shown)


andrix10 (approved):

danielakhterov:
Merge as this is an important fix that needs to be released ASAP. However, I'd still like any reviewers to thoroughly review my changes in case I've made more silly mistakes.

topic_message_query.go
```diff
@@ -115,77 +108,81 @@ func (query *TopicMessageQuery) Subscribe(client *Client, onNext func(TopicMessa
...
+				} else if err == io.EOF {
+					query.completionHandler()
+					break
+				} else {
+					panic(err)
 				}
+			}
 
+			if subClient == nil {
 				ctx, cancel := context.WithCancel(context.TODO())
```
xin-hedera, line 132: `cancel` should be called when a subscription finishes in the goroutine, regardless of success/failure

PR #175, opened 2021-06-16
feat: add `sign-on-demand` functionality

danielakhterov:
Closes: #185

### DIFF
--- a/transaction.go
+++ b/transaction.go
@@ -3,6 +3,7 @@ package hedera
 import (
 	"bytes"
 	"crypto/sha512"
+
 	"github.com/pkg/errors"
 
 	"time"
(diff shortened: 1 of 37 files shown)


transaction.go
```diff
@@ -489,18 +514,62 @@ func (transaction *Transaction) ToBytes() ([]byte, error) {
 
 }
 
-func (transaction *Transaction) buildTransactions(untilIndex int) error {
-	for i := len(transaction.transactions); i < untilIndex; i++ {
-		data, err := protobuf.Marshal(transaction.signedTransactions[i])
+func (transaction *Transaction) signTransaction(index int) {
+	if len(transaction.signedTransactions[index].SigMap.SigPair) != 0 {
```
andrix10, line 518: should probably check if the index is outside bounds of the array here, I am not a fan of those out of bounds errors

andrix10 (approved):

transaction.go
danielakhterov: Yea, my reasoning here is that since this an internal API if we somehow provide an invalid index we shouldn't be checking that, but instead it should panic.

PR #176, opened 2021-06-16
feat: added parsing file 0.0.111 support

### LINKED ISSUE #174: Support parsing file `0.0.111`

andrix10:
Closes #174 

### DIFF
--- a/fee_components.go
+++ b/fee_components.go
@@ -1,0 +1,88 @@
+package hedera
+
+import (
+	"fmt"
+	protobuf "github.com/golang/protobuf/proto"
+	"github.com/hashgraph/hedera-sdk-go/v2/proto"
+)
+
+type FeeComponents struct {
+	Min                        int64
+	Max                        int64
+	Constant                   int64
+	TransactionBandwidthByte   int64
+	TransactionVerification    int64
+	TransactionRamByteHour     int64
+	TransactionStorageByteHour int64
+	ContractTransactionGas     int64
+	TransferVolumeHbar         int64
+	ResponseMemoryByte         int64
+	ResponseDiscByte           int64
+}
+
+func feeComponentsFromProtobuf(feeComponents *proto.FeeComponents) (FeeComponents, error) {
+	if feeComponents == nil {
+		return FeeComponents{}, errParameterNull
+	}
+
+	return FeeComponents{
+		Min:                        feeComponents.GetMin(),
+		Max:                        feeComponents.GetMax(),
+		Constant:                   feeComponents.GetConstant(),
+		TransactionBandwidthByte:   feeComponents.GetBpt(),
+		TransactionVerification:    feeComponents.GetVpt(),
+		TransactionRamByteHour:     feeComponents.GetRbh(),
+		TransactionStorageByteHour: feeComponents.GetSbh(),
+		ContractTransactionGas:     feeComponents.GetGas(),
+		TransferVolumeHbar:         feeComponents.GetTv(),
+		ResponseMemoryByte:         feeComponents.GetBpr(),
+		ResponseDiscByte:           feeComponents.GetSbpr(),
+	}, nil
+}
+
+func (feeComponents FeeComponents) toProtobuf() *proto.FeeComponents {
+	return &proto.FeeComponents{
+		Min:      feeComponents.Min,
+		Max:      feeComponents.Max,
+		Constant: feeComponents.Constant,
+		Bpt:      feeComponents.TransactionBandwidthByte,
+		Vpt:      feeComponents.TransactionVerification,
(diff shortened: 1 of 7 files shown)


fee_components.go
```diff
@@ -0,0 +1,106 @@
+package hedera
+
+import (
+	"fmt"
+	protobuf "github.com/golang/protobuf/proto"
+	"github.com/hashgraph/hedera-sdk-go/v2/proto"
+)
+
+type FeeComponents struct {
+	Min int64
```
danielakhterov, line 10: Definitely looks like `go fmt` was not run; we need to add that to CI.

fee_components.go
```diff
@@ -0,0 +1,106 @@
...
+	TransactionVerification int64
+	TransactionRamByteHour int64
+	TransactionStorageByteHour int64
+	ContractTransactionGas int64
+	TransferVolumeHbar int64
+	ResponseMemoryByte int64
+	ResponseDiscByte int64
+}
+
+func newFeeComponents(min int64, max int64, constant int64, transactionBandwidthByte int64, transactionVerification int64,
```
danielakhterov, line 23: I don't think there is a point for this to exist really. All the variables here are `int64` and the default value is `0`. Might as well construct the type using the struct initialization notation.

fee_components.go
```diff
@@ -0,0 +1,106 @@
...
+	info, err := feeComponentsFromProtobuf(&pb)
+	if err != nil {
+		return FeeComponents{}, err
+	}
+
+	return info, nil
+}
+
+func (feeComponents FeeComponents) String() string {
+	return fmt.Sprintf("Min: %d, Max: %d, Constant: %d,TransactionBandwithByte: %d,TransactionVerification: %d,TransactionRamByteHour: %d,TransactionStorageByteHour: %d, ContractTransactionGas: %d,TransferVolumeHbar: %d, ResponseMemoryByte: %d, ResponseDiscByte: %d", feeComponents.Min, feeComponents.Max, feeComponents.Constant, feeComponents.TransactionBandwidthByte, feeComponents.TransactionVerification, feeComponents.TransactionRamByteHour, feeComponents.TransactionStorageByteHour, feeComponents.ContractTransactionGas, feeComponents.TransferVolumeHbar, feeComponents.ResponseMemoryByte, feeComponents.ResponseDiscByte)
```
danielakhterov, line 105: I'm wondering if we can simply use `fmt.Sprintf("%+v\n", feeComponents)` instead here?

fee_schedules_test.go
```diff
@@ -0,0 +1,18 @@
...
+func TestFetchFeeSchedules_Execute(t *testing.T) {
+	env := NewIntegrationTestEnv(t)
+
+	feeSchedulesBytes, err := NewFileContentsQuery().
+		SetFileID(FileID{Shard: 0, Realm: 0, File:  111}).
+		Execute(env.Client)
+	assert.NoError(t, err)
+	feeSchedules, err := FeeSchedulesFromBytes(feeSchedulesBytes)
+	assert.NoError(t, err)
+	assert.NotNil(t, feeSchedules)
```
danielakhterov, line 17: Can you add some checks after this the values of `feeSchedules` aren't all `0`?

request_type.go
```diff
@@ -0,0 +1,203 @@
...
+import (
+	"fmt"
+)
+
+type RequestType uint32
+
+
+
+const (
+	RequestTypeNONE           RequestType = 0 // UNSPECIFIED - Need to keep first value as unspecified because first element is ignored and not parsed (0 is ignored by parser)
```
danielakhterov, line 12: Unlike the protobufs, lets be consistent. All enum variants should be `PascalCase` so `RequestTypeNone`.
Also, can you move all the comments above each enum instead of after it. I don't think we use the comments after code anywhere in the code base.

request_type.go
```diff
@@ -0,0 +1,203 @@
...
+		return "SCHEDULE_CREATE"
+	case RequestTypeScheduleDelete:
+		return "SCHEDULE_DELETE"
+	case RequestTypeScheduleSign:
+		return "SCHEDULE_SIGN"
+	case RequestTypeScheduleGetInfo:
+		return "SCHEDULE_GET_INFO"
+	}
+
+	panic(fmt.Sprintf("unreacahble: Status.String() switch statement is non-exhaustive. RequestType: %v", uint32(requestType)))
```
danielakhterov, line 202: `Status.String()` should be `RequestType.String()`

danielakhterov (changes requested):

fee_components.go
andrix10: Not possible it seems, it errors out with `golang causes recursive String method call`

danielakhterov (approved):

PR #179, opened 2021-06-17
feat: add linting check to CI

danielakhterov:
Resolves: #178


mehcode:
https://github.com/golangci/golangci-lint is the lint tool we should be using in CI

PR #181, opened 2021-06-17
Feature/id checksum

### LINKED ISSUE #180: Entity ID Checksums

andrix10:
Closes #180


account_id.go
```diff
@@ -3,30 +3,57 @@ package hedera
...
+	var networkNames = []NetworkName{
+		Mainnet,
+		Testnet,
+		Previewnet,
+	}
+
+	var network NetworkName
+	for _, name := range networkNames {
+		checksum, err = checksumParseAddress(name.Network(), data)
+		if err != nil {
```
danielakhterov, line 35: You're not supposed to err here. You're supposed to try the next network name. If all 3 fail, then you err.

Actually, maybe I'm misunderstanding. Not sure what's the difference between err and `checksum.status`

account_id.go
```diff
@@ -3,30 +3,57 @@ package hedera
...
+			break
+		}
+	}
+
+	err = checksumVerify(checksum.status)
 	if err != nil {
 		return AccountID{}, err
 	}
 
+	tempChecksum := checksum.correctChecksum
```
danielakhterov, line 49: Is there a point to this variable? Can you not assign `checksum.correctChecksum` direct to `AccountId.Checksum`?

account_id.go
```diff
@@ -39,16 +66,59 @@ func AccountIDFromSolidityAddress(s string) (AccountID, error) {
...
-		Account: account,
+		Shard:    shard,
+		Realm:    realm,
+		Account:  account,
+		Checksum: nil,
+		Network:  nil,
 	}, nil
 }
 
+func (id *AccountID) GetNetworkFromChecksum() error {
```
danielakhterov, line 77: Should not be public.

account_id.go
```diff
@@ -39,16 +66,59 @@ func AccountIDFromSolidityAddress(s string) (AccountID, error) {
...
+			network = name
+			break
+		}
+	}
+
+	id.Network = &network
+	return nil
+}
+
+func AccountIDValidateNetworkOnIDs(id AccountID, other AccountID) error {
```
danielakhterov, line 106: Should not be public.

account_id.go
```diff
@@ -39,16 +66,59 @@ func AccountIDFromSolidityAddress(s string) (AccountID, error) {
...
+	}
+
+	return nil
+}
+
 // String returns the string representation of an AccountID in
 // `Shard.Realm.Account` (for example "0.0.3")
 func (id AccountID) String() string {
-	return fmt.Sprintf("%d.%d.%d", id.Shard, id.Realm, id.Account)
+	checksum, err := checksumParseAddress("", fmt.Sprintf("%d.%d.%d", id.Shard, id.Realm, id.Account))
```
danielakhterov, line 117: We should use the checksum already on the ID; checksums should be calculated when the ID is created.

danielakhterov (changes requested):

PR #201, opened 2021-07-01
fix: fixed legacyDerive

andrix10:


crypto.go
```diff
@@ -265,29 +265,34 @@ func deriveChildKey(parentKey []byte, chainCode []byte, index uint32) ([]byte, [
...
-
-	if index >= 0 {
-		for i := len(parentKey); i < len(parentKey)+4; i++ {
-			password[i] = 0
+func deriveLegacyChildKey(parentKey []byte, index int64) []byte {
+	in := make([]uint8, 8)
+	if index == int64(0xffffffffff) {
+		in[3] = uint8(0xff)
+		for i := 4; i < len(in); i++ {
+			in[i] = uint8(uint64(index) >> 32)
```
danielakhterov, line 273: A bit confused on what this is trying to do. if `index` is `0xff_ffffffff` then you can just set `in` to be `{0, 0, 0, 0xff, 0xff, 0xff, 0xff, 0xff}`?

crypto.go
```diff
@@ -265,29 +265,34 @@ func deriveChildKey(parentKey []byte, chainCode []byte, index uint32) ([]byte, [
...
 		}
 	} else {
-		for i := len(parentKey); i < len(parentKey)+4; i++ {
-			password[i] = 0xFF
+		if index < 0 {
+			for i := 0; i < 4; i++ {
+				in[i] = uint8(0xff)
+			}
+		} else {
+			for i := 0; i < 4; i++ {
```
danielakhterov, line 281: No point in doing this since `make(...)` will default initialize to zero.

crypto.go
```diff
@@ -364,8 +369,8 @@ func (sk PrivateKey) Derive(index uint32) (PrivateKey, error) {
 	return derivedKey, nil
 }
 
-func (sk PrivateKey) LegacyDerive(index int32) (PrivateKey, error) {
-	keyData := deriveLegacyChildKey(sk.keyData, index)
+func (sk PrivateKey) LegacyDerive(index int64) (PrivateKey, error) {
```
danielakhterov, line 372: This is technically a breaking change. Although I do understand we kind of need to do this. Hmm....

andrix10: if it was java would be easier here

danielakhterov (approved):

PR #206, opened 2021-07-08
feat: improving test env to return unused Hbar

### LINKED ISSUE #205: Delete created accounts after integration tests finish
### Problem

Right now we create an account, and forget about it after each test.

### Solution

We should delete the account (and token if needed) before we finish the test so the main account running the tests doesn't lose as many Hbars per run.

### Alternatives

_No response_

andrix10:

Closes #205

### DIFF
--- a/account_balance_query_test.go
+++ b/account_balance_query_test.go
@@ -9,10 +9,21 @@ import (
 func TestAccountBalanceQuery_Execute(t *testing.T) {
 	env := NewIntegrationTestEnv(t)
 
-	_, err := NewAccountBalanceQuery().
+	balance, err := NewAccountBalanceQuery().
+		SetAccountID(env.OriginalOperatorID).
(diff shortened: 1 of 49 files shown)


account_balance_query_test.go
```diff
@@ -9,11 +9,22 @@ import (
...
+
+	println(balance.Hbars.String())
+
+	_, err = NewAccountBalanceQuery().
 		SetAccountID(env.OperatorID).
 		SetNodeAccountIDs(env.NodeAccountIDs).
 		Execute(env.Client)
 	assert.NoError(t, err)
+
+	err = CloseIntegrationTestEnv(env, nil)
```
danielakhterov, line 26: I'd recommend you simply pass in `t` into `CloseIntegrationTestEnv` so it can automatically do the check.

danielakhterov (approved):

Issue #211, opened 2021-07-15
nil pointer dereference in token fee schedule update


xin-hedera:
https://github.com/hashgraph/hedera-sdk-go/blob/e11efc8058ee5f08d10fec6bb1463499845ca19c/token_fee_schedule_update_transaction.go#L88

The denominating token ID of a fixed fee can be nil (HBAR).

PR #212, opened 2021-07-16
fix: nfts and custom fees

andrix10:

Closes: #211


xin-hedera:
This doesn't seem to fix #211. We also need a test case to capture the nil pointer deference.

andrix10:
> This doesn't seem to fix #211. We also need a test case to capture the nil pointer deference.

yep, it seems i pushed the wrong branch, should definately prune some repos

PR #223, opened 2021-07-24
refactor: clean up client creation in examples

andrix10:

Closes: #216

### DIFF
--- a/examples/consensus_pub_sub/main.go
+++ b/examples/consensus_pub_sub/main.go
@@ -14,34 +14,25 @@ func main() {
 	var client *hedera.Client
 	var err error
 
-	if os.Getenv("HEDERA_NETWORK") == "previewnet" {
-		client = hedera.ClientForPreviewnet()
-	} else {
-		client, err = hedera.ClientFromConfigFile(os.Getenv("CONFIG_FILE"))
-
-		if err != nil {
-			client = hedera.ClientForTestnet()
-		}
+	client, err = hedera.ClientForName(os.Getenv("HEDERA_NETWORK"))
+	if err != nil {
+		println(err.Error(), ": error creating client")
+		return
 	}
 
-	configOperatorID := os.Getenv("OPERATOR_ID")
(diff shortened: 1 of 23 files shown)


examples/consensus_pub_sub/main.go
```diff
@@ -14,14 +14,11 @@ func main() {
...
-	if os.Getenv("HEDERA_NETWORK") == "previewnet" {
-		client = hedera.ClientForPreviewnet()
-	} else {
-		client, err = hedera.ClientFromConfigFile(os.Getenv("CONFIG_FILE"))
-
-		if err != nil {
-			client = hedera.ClientForTestnet()
-		}
+	net := os.Getenv("HEDERA_NETWORK")
+	client, err = hedera.ClientForName(net)
```
danielakhterov, line 18: Could simplify this a bit to `client, err := hedera.ClientForName(os.Getenv("HEDERA_NETWORK"))`

examples/consensus_pub_sub/main.go
```diff
@@ -14,14 +14,11 @@ func main() {
...
+	if err != nil {
+		println(err.Error(), ": error creating client")
+		return
 	}
 
 	configOperatorID := os.Getenv("OPERATOR_ID")
```
danielakhterov, line 24: Remove these variables and the conditional 
```go
	if configOperatorID != "" && configOperatorKey != "" && client.GetOperatorPublicKey().Bytes() == nil {
```
Instead do
```go
		operatorAccountID, err := hedera.AccountIDFromString(os.Getenv("OPERATOR_ID"))
		if err != nil {
			println(err.Error(), ": error converting string to AccountID")
			return
		}
```
Do the same for `OPERATOR_KEY`

danielakhterov (changes requested):

examples/consensus_pub_sub/main.go
andrix10: i see even more cleaner

danielakhterov (approved):

Issue #228, opened 2021-07-30
Create a code generation tool such as `stringer` which generates all the setters and getters


danielakhterov:
### Problem

Much like Java, Go has *a lot* of noise that is due to setters and getters.

### Solution

If we develop some code generation tool to automatically generate those methods using `go generate` we can eliminate this noise hence making the SDK more readable.

PR #231, opened 2021-08-03
feat(wip): implement a generator which generates all the getters, set…

danielakhterov:
**Description**:

**Related issue(s)**:

Closes: #228 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/account_balance_query.go
+++ b/account_balance_query.go
@@ -59,7 +59,7 @@ func (query *AccountBalanceQuery) GetContractID() ContractID {
 	return query.contractID
 }
 
-func (query *AccountBalanceQuery) validateNetworkOnIDs(client *Client) error {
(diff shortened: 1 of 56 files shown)


danielakhterov:
@xin-hedera @andrix10 @tomzhenghedera I'd like your feedback here. The objective of this PR was to auto generate all the getters/setters, from/to protobuf methods, and automatically generate add all the signing methods. Right now it only tests `AccountCreateTransaction`, but the point is this would be applied to all requests. When I first started working on this I was heavily in favor of just generating nearly everything since it'd allow us to move forward with new requests and fix bugs faster. However, since then I've had some conversations about this which strongly go against this amount of code generation. I'm a bit on the fence about if we should actually use this PR at all or just scrap it, so I'm wondering what you guys' think.

PR #232, opened 2021-08-04
feat: adding Client.pingAll(), Client will now use validated nodes for…

andrix10:
… session run

Closes: #183

### DIFF
--- a/account_balance.go
+++ b/account_balance.go
@@ -14,11 +14,17 @@ type AccountBalance struct {
 	TokenDecimals TokenDecimalMap
 }
 
-func accountBalanceFromProtobuf(pb *proto.CryptoGetAccountBalanceResponse) AccountBalance {
-	tokens := make(map[TokenID]uint64, len(pb.TokenBalances))
-	for _, token := range pb.TokenBalances {
-		t := tokenIDFromProtobuf(token.TokenId)
-		tokens[t] = token.Balance
+func accountBalanceFromProtobuf(pb *proto.CryptoGetAccountBalanceResponse, networkName *NetworkName) AccountBalance {
+	if pb == nil {
+		return AccountBalance{}
+	}
+	var tokens map[TokenID]uint64
+	if pb.TokenBalances != nil {
+		tokens = make(map[TokenID]uint64, len(pb.TokenBalances))
+		for _, token := range pb.TokenBalances {
+			t := tokenIDFromProtobuf(token.TokenId, nil)
+			tokens[t] = token.Balance
+		}
 	}
 
 	return AccountBalance{
--- a/network.go
+++ b/network.go
@@ -1,20 +1,28 @@
 package hedera
 
 import (
+	"math"
 	"sort"
+	"time"
 )
 
 type network struct {
-	network      map[string]AccountID
-	nodes        []*node
-	networkNodes map[AccountID]*node
+	network                map[string]AccountID
+	nodes                  []*node
+	networkNodes           map[AccountID]*node
+	maxNodeAttempts        int
+	nodeWaitTime           time.Duration
+	maxNodesPerTransaction *int
 }
 
 func newNetwork() network {
 	return network{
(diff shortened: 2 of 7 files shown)


network.go
```diff
@@ -52,6 +60,27 @@ func (network *network) SetNetwork(net map[string]AccountID) error {
...
+				nod = network.nodes[i]
+			} else {
+				continue
+			}
+			if nod.attempts >= int64(network.maxNodeAttempts) {
+				err := nod.close()
+				if err != nil {
+					panic(err)
+				}
+				network.nodes = append(network.nodes[:i], network.nodes[i+1:]...)
```
danielakhterov, line 76: I just realized that this is how you're supposed to delete elements from a list in Go :vomiting_face:

node.go
```diff
@@ -69,9 +87,20 @@ func (node *node) getChannel() (*channel, error) {
...
 	}
 
-	conn, err := grpc.Dial(node.address, grpc.WithInsecure(), grpc.WithKeepaliveParams(kacp), grpc.WithBlock())
+	var conn *grpc.ClientConn
+	var err error
+	parts := strings.SplitN(node.address, ":", 2)
+	if parts[1] == "443" || parts[1] == "50212" {
+		cont, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+		defer cancel()
+		conn, err = grpc.DialContext(cont, node.address, grpc.WithTransportCredentials(credentials.NewTLS(&tls.Config{InsecureSkipVerify: true})), grpc.WithKeepaliveParams(kacp), grpc.WithBlock())
```
danielakhterov, line 96: Instead of making a conditional and calling `grpc.DialContext()` context you can move that to outside the conditional and instead only save the security inside the conditional.
```go
security := grpc.WithInsecure()

if parts[1] == "443" || parts[1] == "50212" {
    security = grpc.WithTransportCredentials(credentials.NewTLS(&tls.Config{InsecureSkipVerify: true}))
}

conn, err := grpc.Dial(cont, node.address, security, grpc.WithKeepaliveParams(kacp), grpc.WithBlock())
if err != nil {
    panic(err)
}
```

node.go
```diff
@@ -69,9 +87,20 @@ func (node *node) getChannel() (*channel, error) {
 		PermitWithoutStream: true,
 	}
 
-	conn, err := grpc.Dial(node.address, grpc.WithInsecure(), grpc.WithKeepaliveParams(kacp), grpc.WithBlock())
+	var conn *grpc.ClientConn
+	var err error
+	parts := strings.SplitN(node.address, ":", 2)
+	if parts[1] == "443" || parts[1] == "50212" {
+		cont, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+		defer cancel()
```
danielakhterov, line 95: I could be wrong, but I think calling `defer cancel()` will call `cancel` as soon as this variable goes out of scope of the function which would immediately cancel the connection.

danielakhterov (changes requested):

account_balance.go
```diff
@@ -15,10 +15,16 @@ type AccountBalance struct {
...
 
 func accountBalanceFromProtobuf(pb *proto.CryptoGetAccountBalanceResponse, networkName *NetworkName) AccountBalance {
-	tokens := make(map[TokenID]uint64, len(pb.TokenBalances))
-	for _, token := range pb.TokenBalances {
-		t := tokenIDFromProtobuf(token.TokenId, nil)
-		tokens[t] = token.Balance
+	if pb == nil {
+		return AccountBalance{}
+	}
+	var tokens map[TokenID]uint64
```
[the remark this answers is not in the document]
andrix10: wow, alright

node.go
andrix10: i am not sure, but if i don't have it, it will not stop running even if timeout is reached for some reason

danielakhterov (approved):

PR #238, opened 2021-08-05
refactor: id checksums, implement Client.autoValidateChecksums, false…

andrix10:
… by default

Closes: #222

### DIFF
--- a/account_id.go
+++ b/account_id.go
@@ -59,7 +59,7 @@ func AccountIDFromSolidityAddress(s string) (AccountID, error) {
 }
 
 func (id *AccountID) Validate(client *Client) error {
-	if !id.isZero() && client != nil && client.networkName != nil {
+	if !id.isZero() && client != nil && client.networkName != nil{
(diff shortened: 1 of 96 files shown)


account_id.go
```diff
@@ -59,7 +59,7 @@ func AccountIDFromSolidityAddress(s string) (AccountID, error) {
 }
 
 func (id *AccountID) Validate(client *Client) error {
-	if !id.isZero() && client != nil && client.networkName != nil {
+	if !id.isZero() && client != nil && client.networkName != nil && client.autoValidateChecksums {
```
danielakhterov, line 62: `client.autoValidateChecksums` should not be part of the `*ID.Validate()` methods. The point of `*ID.Validate()` methods is to allow the user to force validation of a certain ID. `Client.autoValidateChecksums` on the other hand is strictly meant set if `*.Execute()` should validate checksums of all IDs or not.

danielakhterov (changes requested):

account_id.go
andrix10: oooo

danielakhterov (approved):

PR #240, opened 2021-08-06
feat: remove the use of pb in all requests and store the fields on th…

andrix10:
…e requests structs themselves.


account_balance_query.go
```diff
@@ -76,52 +67,50 @@ func (query *AccountBalanceQuery) validateNetworkOnIDs(client *Client) error {
...
 
-	query.build()
-
 	resp, err := execute(
 		client,
 		request{
 			query: &query.Query,
 		},
 		accountBalanceQuery_shouldRetry,
-		costQuery_makeRequest,
```
danielakhterov, line 124: Shouldn't this be `costQueryMakeRequest`?

andrix10: was  but confused what you meant but, 
`protoReq, err := query.costQueryMakeRequest(client)`

it returns an error so you have to catch it, right

PR #247, opened 2021-08-17
feat: adding royalty fee, updating protobufs

andrix10:

Closes: #245


fee_assessment_method.go
```diff
@@ -0,0 +1,24 @@
+package hedera
+
+type FeeAssessmentMethod bool
+
+const (
+	Inclusive FeeAssessmentMethod = false
```
danielakhterov, line 6: Since this is Go, you'll need to preface the name of the variable with `FeeAssessmentMethod` much like we do with `StatusOk`

fee_assessment_method.go
```diff
@@ -0,0 +1,24 @@
...
+
+func (assessment FeeAssessmentMethod) String() string {
+	if assessment == true {
+		return "Exclusive"
+	}
+
+	return "Inclusive"
+}
+
+func (assessment FeeAssessmentMethod) ValueOf(assess bool) FeeAssessmentMethod {
```
danielakhterov, line 18: This should be private.

custom_fractional_fee.go
```diff
@@ -7,19 +7,21 @@ import (
...
-		Numerator:     fractionalFee.FractionalAmount.Numerator,
-		Denominator:   fractionalFee.FractionalAmount.Denominator,
-		MinimumAmount: fractionalFee.MinimumAmount,
-		MaximumAmount: fractionalFee.MaximumAmount,
+		CustomFee:        fee,
+		Numerator:        fractionalFee.FractionalAmount.Numerator,
+		Denominator:      fractionalFee.FractionalAmount.Denominator,
+		MinimumAmount:    fractionalFee.MinimumAmount,
+		MaximumAmount:    fractionalFee.MaximumAmount,
+		AssessmentMethod: FeeAssessmentMethod(fractionalFee.NetOfTransfers),
```
danielakhterov, line 24: Shouldn't this  use `valueOf`? If this works without using `valueOf` then we should just remove that function.

danielakhterov (changes requested):

danielakhterov (approved):

Issue #246, opened 2021-08-16
Regression - TransferTransaction.Freeze() nil pointer panic


xin-hedera:
https://github.com/hashgraph/hedera-sdk-go/blob/7fbc09e63a5f2470682b51c507ddb17065f0beba/transfer_transaction.go#L428

https://github.com/hashgraph/hedera-sdk-go/blob/7fbc09e63a5f2470682b51c507ddb17065f0beba/transfer_transaction.go#L436

Looks like it regressed for all *Transaction.Freeze()

PR #248, opened 2021-08-17
fix: TransferTransaction.Freeze() nil pointer panic, implemented vali…

andrix10:
…dateNetworkOnIDs nil client check

Closes: #246

### DIFF
--- a/account_create_transaction.go
+++ b/account_create_transaction.go
@@ -142,7 +142,7 @@ func (transaction *AccountCreateTransaction) GetAccountMemo() string {
 }
 
 func (transaction *AccountCreateTransaction) validateNetworkOnIDs(client *Client) error {
-	if !client.autoValidateChecksums {
+	if client == nil || !client.autoValidateChecksums {
 		return nil
 	}
 	return transaction.proxyAccountID.Validate(client)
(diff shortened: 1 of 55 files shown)


account_create_transaction.go
```diff
@@ -142,6 +142,9 @@ func (transaction *AccountCreateTransaction) GetAccountMemo() string {
 }
 
 func (transaction *AccountCreateTransaction) validateNetworkOnIDs(client *Client) error {
+	//if client == nil {
```
danielakhterov, line 145: I think you forgot to un-comment this.

danielakhterov (changes requested):

xin-hedera:
Can we add unit tests?

topic_message_submit_transaction.go
```diff
@@ -127,6 +127,9 @@ func (transaction *TopicMessageSubmitTransaction) SignWith(
 }
 
 func (transaction *TopicMessageSubmitTransaction) validateNetworkOnIDs(client *Client) error {
+	if client == nil {
+		return nil
+	}
 	if !client.autoValidateChecksums {
```
steven-sheehy, line 133: Can't these be combined for succinctness?
```suggestion
	if client == null || !client.autoValidateChecksums {
```

mehcode: First mistake is trying to be succinct in Go. :laughing:

danielakhterov (approved):

danielakhterov:
Created issue for unit tests: #255

PR #259, opened 2021-08-20
feature: implemented Client.[min|max]Backoff()

andrix10:

Closes: #254


client.go
```diff
@@ -285,6 +290,32 @@ func (client *Client) GetNetwork() map[string]AccountID {
 	return client.network.network
 }
 
+func (client *Client) SetMaxBackoff(max time.Duration) {
```
danielakhterov, line 293: This should return `*Client` much like the other set methods.

executable.go
```diff
@@ -13,6 +13,9 @@ import (
 	"google.golang.org/grpc/status"
 )
 
+var minBackoff *time.Duration
```
danielakhterov, line 16: Definitely shouldn't be using globals here. These should be passed into the function.

executable.go
```diff
@@ -177,3 +188,36 @@ func executableDefaultRetryHandler(err error) bool {
 		return false
 	}
 }
+func SetMaxBackoffForExecute(max time.Duration) {
```
danielakhterov, line 191: This method should exist on each request, not in `executable.go`. So `*[Transaction|Query].Set[Min|Max]Backoff()` where the setters return the request.

danielakhterov (changes requested):

client.go
andrix10: Most others don't though

PR #261, opened 2021-08-21
feature: update protobufs

andrix10:


danielakhterov (approved):
Ran integration tests locally and most passed. Reason "most" is okay is because the tests that failed don't seem to be because of the protobuf changes.

Issue #265, opened 2021-08-26
`Client.Set*` methods don't return `*Client`


danielakhterov:
### Problem

`Client` setters should utilize the builder pattern, but do not.

### Solution

Update set methods on `Client` to return `*Client`

PR #267, opened 2021-08-27
fix: Client set* methods now return *Client

andrix10:

Closes: #265

### DIFF
--- a/client.go
+++ b/client.go
@@ -125,7 +125,7 @@ func newClient(network map[string]AccountID, mirrorNetwork []string, name Networ
 		maxBackoff:            8 * time.Second,
 	}
 
-	_ = client.SetNetwork(network)
+	_, _ = client.SetNetwork(network)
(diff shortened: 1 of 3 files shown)


client.go
```diff
@@ -282,42 +282,49 @@ func (client *Client) Close() error {
 
 // SetNetwork replaces all nodes in the Client with a new set of nodes.
 // (e.g. for an Address Book update).
-func (client *Client) SetNetwork(network map[string]AccountID) error {
-	return client.network.SetNetwork(network)
+func (client *Client) SetNetwork(network map[string]AccountID) (*Client, error) {
```
danielakhterov, line 285: Since this method already returned _something_ we can't change the return type :(
You can only update the return type of methods that returned nothing before.

danielakhterov (changes requested):

danielakhterov:
Closing as I completely forgot changing function signature at all is a breaking changes since people might use function pointers which require defining the return type.

PR #323, opened 2021-11-16
fix: get receipt on erroneous executed tx

### LINKED ISSUE #316: Identical Scheduled Transactions - How to get their ScheduleID
From some version of the SDK, when you submit an identical scheduled transaction and you get its receipt, it resolves with `exceptional receipt status: IDENTICAL_SCHEDULE_ALREADY_CREATED`. How exactly do you now have to get the ScheduleID of the original transaction, since the whole receipt is empty?

failfmi:
**Description**:
* When `ScheduleCreate` resolves with `IDENTICAL_SCHEDULE_ALREADY_CREATED`, the receipt should be filled with the original tx data.
* I do not why, but before the receipt was in `resp.query.GetTransactionGetReceipt().GetReceipt()` and now `resp.query` is null. 
* Previously, the receipt on ScheduleCreate **never** resolved with an error, now the case is a bit different. (by previosuly I mean `v2.1.5-beta.3`)
* I've added an example, creating 3 accounts, each of their key being from a threshold 2/3 account, from which a schedule create transaction is executed, when all the signatures are provided.

**Related issue(s)**:

Fixes #316 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/transaction_receipt_query.go
+++ b/transaction_receipt_query.go
@@ -137,7 +137,7 @@ func _TransactionReceiptQueryShouldRetry(request _Request, response _Response) _
 	switch Status(response.query.GetTransactionGetReceipt().GetReceipt().GetStatus()) {
 	case StatusBusy, StatusUnknown, StatusOk, StatusReceiptNotFound, StatusRecordNotFound:
 		return executionStateRetry
-	case StatusSuccess:
+	case StatusSuccess, StatusIdenticalScheduleAlreadyCreated:
 		return executionStateFinished
 	default:
 		return executionStateError
@@ -155,8 +155,7 @@ func _TransactionReceiptQueryMapStatusError(request _Request, response _Response
 	}
 
 	return ErrHederaReceiptStatus{
-		Status:  Status(response.query.GetTransactionGetReceipt().GetReceipt().GetStatus()),
-		Receipt: _TransactionReceiptFromProtobuf(response.query.GetTransactionGetReceipt().GetReceipt()),
+		Status: Status(response.query.GetTransactionGetReceipt().GetReceipt().GetStatus()),
 	}
 }
 
@@ -270,13 +269,10 @@ func (query *TransactionReceiptQuery) Execute(client *Client) (TransactionReceip
 	)
 
 	if err != nil {
-		if precheckErr, ok := err.(ErrHederaPreCheckStatus); ok {
+		switch precheckErr := err.(type) {
+		case ErrHederaPreCheckStatus:
 			return TransactionReceipt{}, _NewErrHederaReceiptStatus(precheckErr.TxID, precheckErr.Status)
 		}
-		if respErr, ok := err.(ErrHederaReceiptStatus); ok {
-			return respErr.Receipt, nil
-		}
-
 		return TransactionReceipt{}, err
 	}
 
--- a/transaction_receipt_query.go
+++ b/transaction_receipt_query.go
@@ -271,17 +271,12 @@ func (query *TransactionReceiptQuery) Execute(client *Client) (TransactionReceip
 
 	if err != nil {
 		if precheckErr, ok := err.(ErrHederaPreCheckStatus); ok {
-			receipt := TransactionReceipt{}
-			if resp.query.GetTransactionGetReceipt() != nil {
-				receipt = _TransactionReceiptFromProtobuf(resp.query.GetTransactionGetReceipt().GetReceipt())
-			}
-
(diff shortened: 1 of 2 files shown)


transaction_receipt_query.go
```diff
@@ -271,17 +271,12 @@ func (query *TransactionReceiptQuery) Execute(client *Client) (TransactionReceip
...
-			}
-
-			return receipt, ErrHederaReceiptStatus{
-				TxID:    precheckErr.TxID,
-				Status:  precheckErr.Status,
-				Receipt: receipt,
-			}
+			return TransactionReceipt{}, _NewErrHederaReceiptStatus(precheckErr.TxID, precheckErr.Status)
 		}
+		if respErr, ok := err.(ErrHederaReceiptStatus); ok {
```
danielakhterov, line 276: `_Execute` should never return an `ErrHederaReceiptStatus` and if it does, that would likely be the bug that needs to be fixed.

transaction_receipt_query.go
```diff
@@ -271,17 +271,12 @@ func (query *TransactionReceiptQuery) Execute(client *Client) (TransactionReceip
...
+			return respErr.Receipt, nil
```
danielakhterov, line 277: In the result of an error we should return both, a filled out receipt and an error.

danielakhterov (changes requested):

transaction_receipt_query.go
failfmi: Yes, this is happening.

transaction_receipt_query.go
failfmi: In my opinion, getting a transaction's receipt should not resolve with an error, just because the transaction did not result with SUCCESS. An error would be something that happens when you fail to get the receipt (e.g. timeout rpc call). I think that two things are mixed here  - having a non SUCCESSFUL transaction and an actual error while trying to get the receipt.

transaction_receipt_query.go
failfmi: You can use the example I have provided to test it.

danielakhterov: To clarify I was saying we should probably fix that issue in this PR instead of checking for `ErrHederaReceiptStatus`, or at least we should make another PR that fixes that issue.

danielakhterov: Looked a bit at the code, fix _should(?)_ be as simple as changing `_TransactionReceiptQueryMapStatusError` to return `ErrHederaPreCheckStatus` instead of `ErrHederaReceiptStatus`

failfmi: Why would you want to return `ErrHederaPreCheckStatus` given that the error is not from this type?

transaction_receipt_query.go
failfmi: From a developer standpoint, if this return an error, I have to always check that the error is `exceptional receipt status: IDENTICAL_SCHEDULE_ALREADY_CREATED`, so that I am sure that the receipt is filled.

danielakhterov: Yea, that sounds fair.

transaction_receipt_query.go
danielakhterov: The reason is for the same reason you've mentioned before in some other comment .`TransactionReceiptQuery` itself should not error on anything except when `Header.NodeTransactionPrecheckCode` is an erring value; `Receipt.Status` should not be used to determine if `TransactionReceiptQuery` failed. However, we do want to return an error when someone uses the convince methods `Transaction[Response|ID].getReceipt()` as those add the extra requirement of `TransactionReceipt.Status` being `SUCCESS` else the receipt is determined to be an error. At least this is the expected behavior.

transaction_receipt_query.go
```diff
@@ -137,7 +137,7 @@ func _TransactionReceiptQueryShouldRetry(request _Request, response _Response) _
 	switch Status(response.query.GetTransactionGetReceipt().GetReceipt().GetStatus()) {
 	case StatusBusy, StatusUnknown, StatusOk, StatusReceiptNotFound, StatusRecordNotFound:
 		return executionStateRetry
-	case StatusSuccess:
+	case StatusSuccess, StatusIdenticalScheduleAlreadyCreated:
```
danielakhterov, line 140: @failfmi We should be treating all receipt statuses that we don't retry on as successful.
https://github.com/hashgraph/hedera-sdk-go/pull/327

failfmi:
Closing this, as the _correct_ fix has been made in #327..

PR #327, opened 2021-11-18
fix: `TransactionReceiptQuery` should not error if a receipt is present

danielakhterov:
Closes: #316

### DIFF
--- a/examples/schedule_identical_transaction/main.go
+++ b/examples/schedule_identical_transaction/main.go
@@ -1,0 +1,218 @@
+package main
+
+import (
+	"fmt"
+	"github.com/hashgraph/hedera-sdk-go/v2"
+	"os"
+)
+
+func main() {
+	var client *hedera.Client
+	var err error
+
+	// Retrieving network type from environment variable HEDERA_NETWORK
+	client, err = hedera.ClientForName(os.Getenv("HEDERA_NETWORK"))
+	if err != nil {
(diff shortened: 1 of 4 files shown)


examples/schedule_identical_transaction/main.go
```diff
@@ -0,0 +1,218 @@
...
+		scheduledTx = scheduledTx.
+			SetPayerAccountID(thresholdAccount)
+
+		response, err := scheduledTx.Execute(client)
+		if err != nil {
+			println(err.Error(), ": error while executing schedule create transaction with operator", operator)
+			return
+		}
+
+		receipt, err := hedera.NewTransactionReceiptQuery().
```
danielakhterov, line 159: @failfmi You'd be required to query the receipt using `NewTransactionReceiptQuery` directly instead of `GetReceipt()` since the object of `GetReceipt()` is to return an error if the status is not `SUCCESS`

failfmi: To be honest, I do not understand why would have two behaviours for almost the same thing, but since this is right way - LGTM!

failfmi: Is this somewhere documented?

danielakhterov: Unfortunately it isn't, but we really should.

andrix10 (approved):

PR #340, opened 2021-12-01
feat: adding ECDSA secp256k1 support

andrix10:

Closes: #335


crypto.go
```diff
@@ -36,7 +28,7 @@ func _KeyFromProtobuf(pbKey *services.Key) (Key, error) {
 	}
 	switch key := pbKey.GetKey().(type) {
 	case *services.Key_Ed25519:
-		return PublicKeyFromBytes(key.Ed25519)
+		return Ed25519PublicKeyFromBytes(key.Ed25519)
```
danielakhterov, line 31: We should have such a method as public. Instead we should have `PublicKeyFromBytesEd25519()`. The reason this distinction is necessary is because `Ed25519*` should not be public as it's a private type, but `PublicKey` is public and hence should have public methods for constructing keys from `Ed25519`.

crypto.go
```diff
@@ -59,194 +51,229 @@ func _KeyFromProtobuf(pbKey *services.Key) (Key, error) {
...
 
-// PrivateKeyFromMnemonic recovers an PrivateKey from a valid 24 word length mnemonic phrase and a
-// passphrase.
-//
-// An empty string can be passed for passPhrase If the mnemonic phrase wasn't generated with a passphrase. This is
-// required to recover a private key from a mnemonic generated by the Android and iOS wallets.
-func PrivateKeyFromMnemonic(mnemonic Mnemonic, passPhrase string) (PrivateKey, error) {
-	salt := []byte("mnemonic" + passPhrase)
-	seed := pbkdf2.Key([]byte(mnemonic.String()), salt, 2048, 64, sha512.New)
+func Ed25519PrivateKeyFromBytes(bytes []byte) (PrivateKey, error) {
```
danielakhterov, line 92: Use `PrivateKeyFromBytesEd25519()` instead

crypto.go
```diff
@@ -304,158 +331,310 @@ func _DeriveLegacyChildKey(parentKey []byte, index int64) []byte {
...
+		return pk.ed25519PublicKey._BytesRaw()
+	}
+
+	return []byte{}
 }
 
-// SupportsDerivation returns true if the PrivateKey supports derivation.
-func (sk PrivateKey) SupportsDerivation() bool {
-	return sk.chainCode != nil
+func (pk PublicKey) RawBytes() []byte {
```
danielakhterov, line 470: `BytesRaw()`

crypto.go
```diff
@@ -304,158 +331,310 @@ func _DeriveLegacyChildKey(parentKey []byte, index int64) []byte {
...
+		return sk.ed25519PrivateKey._BytesRaw()
+	}
+
+	return []byte{}
 }
 
-// Keystore returns an encrypted _Keystore containing the PrivateKey.
-func (sk PrivateKey) Keystore(passphrase string) ([]byte, error) {
-	return _NewKeystore(sk.keyData, passphrase)
+func (sk PrivateKey) DerBytes() []byte {
```
danielakhterov, line 434: `BytesDer()`

ecdsa_private_key.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+}
+
+func _RawECDSAPrivateKeyFromBytes(byt []byte) (*_ECDSAPrivateKey, error) {
+	length := len(byt)
+	if length != 32 {
+		return &_ECDSAPrivateKey{}, _NewErrBadKeyf("invalid private key length: %v bytes", len(byt))
+	}
+
+	b := bytes.Trim(byt, "\x00")
```
danielakhterov, line 52: I don't think you should be trimming `\x00`'s from the start. Just `0x`

danielakhterov (changes requested):
Haven't reviewed everything, but can we change the naming scheme to `<type><functionality><version>`. For instance, `_RawECDSAPublicKeyFromBytes` should be `_ECDSAPublicKeyFromBytesRaw()` since `ECDSAPublicKey` is the `<type>`, `FromBytes` is the `<functionality>` and `Raw` is the `<version>` of `FromBytes`

ecdsa_private_key.go
andrix10: I am trimming empty bytes here, because as I saw in how the private key is made it can have empty bytes in order to make it len 32.

crypto.go
```diff
@@ -59,194 +51,229 @@ func _KeyFromProtobuf(pbKey *services.Key) (Key, error) {
...
 
-func PrivateKeyFromPem(bytes []byte, passphrase string) (PrivateKey, error) {
-	var blockType string
-
-	if len(passphrase) == 0 {
-		blockType = "PRIVATE KEY"
-	} else {
-		// the pem is encrypted
-		blockType = "ENCRYPTED PRIVATE KEY"
+func ECDSAPrivateKeyFromString(s string) (PrivateKey, error) {
```
danielakhterov, line 187: Do you mean `PrivateKeyFromStringECDSA()`?

danielakhterov (changes requested):

PR #345, opened 2021-12-08
feat: implementing auto account creation

andrix10:

Closes: #336

### DIFF
--- a/crypto.go
+++ b/crypto.go
@@ -435,15 +435,7 @@ func (sk PrivateKey) PublicKey() PublicKey {
 }
 
 func (sk PrivateKey) ToAccountID(shard uint64, realm uint64) *AccountID {
-	temp := sk.PublicKey()
-
-	return &AccountID{
-		Shard:    shard,
-		Realm:    realm,
-		Account:  0,
-		AliasKey: &temp,
-		checksum: nil,
-	}
+	return sk.PublicKey().ToAccountID(shard, realm)
 }
 
 func (pk PublicKey) ToAccountID(shard uint64, realm uint64) *AccountID {
--- a/token_transfer.go
+++ b/token_transfer.go
@@ -12,6 +12,10 @@ type TokenTransfer struct {
 	Amount    int64
 }
 
+type _TokenTransfers struct {
+	transfers []TokenTransfer
+}
+
 func NewTokenTransfer(accountID AccountID, amount int64) TokenTransfer {
 	return TokenTransfer{
 		AccountID: accountID,
@@ -66,4 +70,39 @@ func TokenTransferFromBytes(data []byte) (TokenTransfer, error) {
 
 func (transfer TokenTransfer) String() string {
 	return fmt.Sprintf("accountID: %s, amount: %d", transfer.AccountID.String(), transfer.Amount)
+}
+
+func (transfers _TokenTransfers) Len() int {
+	return len(transfers.transfers)
+}
+func (transfers _TokenTransfers) Swap(i, j int) {
+	transfers.transfers[i], transfers.transfers[j] = transfers.transfers[j], transfers.transfers[i]
+}
+
+func (transfers _TokenTransfers) Less(i, j int) bool {
+	if transfers.transfers[i].AccountID.Shard < transfers.transfers[j].AccountID.Shard { //nolint
+		return true
+	} else if transfers.transfers[i].AccountID.Shard > transfers.transfers[j].AccountID.Shard {
+		return false
+	}
+
+	if transfers.transfers[i].AccountID.Realm < transfers.transfers[j].AccountID.Realm { //nolint
+		return true
+	} else if transfers.transfers[i].AccountID.Realm > transfers.transfers[j].AccountID.Realm {
+		return false
+	}
+
+	if transfers.transfers[i].AccountID.AliasKey != nil && transfers.transfers[j].AccountID.AliasKey != nil {
+		if transfers.transfers[i].AccountID.String() < transfers.transfers[j].AccountID.String() { //nolint
+			return true
+		} else if transfers.transfers[i].AccountID.String() > transfers.transfers[j].AccountID.String() {
+			return false
+		}
+	}
+
+	if transfers.transfers[i].AccountID.Account < transfers.transfers[j].AccountID.Account { //nolint
+		return true
+	} else { //nolint
+		return false
+	}
 }
--- a/transfer_transaction.go
+++ b/transfer_transaction.go
@@ -102,7 +102,10 @@ func (transaction *TransferTransaction) GetTokenTransfers() map[TokenID][]TokenT
 			})
 		}
 
-		transfers[tokenID] = tokenTransfersList
+		tempTokenTransferList := _TokenTransfers{tokenTransfersList}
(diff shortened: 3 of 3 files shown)


account_id.go
```diff
@@ -112,7 +124,12 @@ func (id AccountID) ToStringWithChecksum(client *Client) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	return fmt.Sprintf("%d.%d.%d-%s", id.Shard, id.Realm, id.Account, checksum.correctChecksum), nil
+
+	if id.AliasKey == nil {
+		return fmt.Sprintf("%d.%d.%d-%s", id.Shard, id.Realm, id.Account, checksum.correctChecksum), nil
+	} else {
+		return fmt.Sprintf("%d.%d.%s-%s", id.Shard, id.Realm, id.AliasKey.String(), checksum.correctChecksum), nil
```
danielakhterov, line 131: We should error here since `checksum` is not supported by `AccountID` with an `AliasKey`

account_id.go
```diff
@@ -126,10 +143,22 @@ func (id AccountID) ToSolidityAddress() string {
 }
 
 func (id AccountID) _ToProtobuf() *services.AccountID {
-	return &services.AccountID{
-		ShardNum:   int64(id.Shard),
-		RealmNum:   int64(id.Realm),
-		AccountNum: int64(id.Account),
+	if id.AliasKey == nil {
```
danielakhterov, line 146: Instead of making two blocks for the entire `AccountID` make an `if` block only for the `AccoutNum` part to de-duplicate at least some of the lines.

account_id.go
```diff
@@ -151,18 +180,36 @@ func _AccountIDFromProtobuf(accountID *services.AccountID) *AccountID {
...
 }
 
 func (id AccountID) _IsZero() bool {
-	return id.Shard == 0 && id.Realm == 0 && id.Account == 0
+	return id.Shard == 0 && id.Realm == 0 && id.Account == 0 && id.AliasKey == nil
 }
 
 func (id AccountID) _Equals(other AccountID) bool {
+	if id.AliasKey != nil && other.AliasKey != nil {
+		return id.Shard == other.Shard && id.Realm == other.Realm && id.Account == other.Account && id.AliasKey.String() == other.AliasKey.String()
```
danielakhterov, line 210: Instead of duplicating the `id.Shard == other.Shard` etc extract `id.AliasKey.String() == other.AliasKey.String()` into a variable and use it in a single return.

account_id.go
```diff
@@ -201,11 +248,19 @@ func (id AccountID) Compare(given AccountID) int {
 		return -1
 	}
 
+	if id.AliasKey != nil && given.AliasKey != nil {
+		if id.AliasKey.String() > given.AliasKey.String() { //nolint
```
danielakhterov, line 252: I believe this is going to compare the pointers to the strings, not the actual values. I believe you need to use `strings.Compare(id.AliasKey.String(), given.AliasKey.String())` instead.

account_id.go
```diff
@@ -230,6 +285,14 @@ func (accountIDs _AccountIDs) Less(i, j int) bool {
 		return false
 	}
 
+	if accountIDs.accountIDs[i].AliasKey != nil && accountIDs.accountIDs[j].AliasKey != nil {
+		if accountIDs.accountIDs[i].String() < accountIDs.accountIDs[j].String() { //nolint
```
danielakhterov, line 289: `strings.Compare()`

account_info.go
```diff
@@ -26,6 +26,7 @@ type AccountInfo struct {
 	AccountMemo                    string
 	OwnedNfts                      int64
 	MaxAutomaticTokenAssociations  uint32
+	AliasKey                       *[]byte
```
danielakhterov, line 29: Should be `*PublicKey`

crypto.go
```diff
@@ -311,6 +311,18 @@ func (sk PrivateKey) PublicKey() PublicKey {
 	}
 }
 
+func (sk PrivateKey) ToAccountId(shard uint64, realm uint64) *AccountID {
```
danielakhterov, line 314: You also want this function on `PublicKey`

id.go
```diff
@@ -8,12 +8,12 @@ import (
 	"strings"
 )
 
-func _IdFromString(s string) (shard int, realm int, num int, checksum *string, err error) {
+func _IdFromString(s string) (shard int, realm int, num int, checksum *string, alias *PublicKey, err error) {
```
danielakhterov, line 11: We really should not support `PublicKey` for all `ID` types

transaction_record.go
```diff
@@ -23,6 +23,7 @@ type TransactionRecord struct {
 	CallResultIsCreate         bool
 	AssessedCustomFees         []AssessedCustomFee
 	AutomaticTokenAssociations []TokenAssociation
+	AliasKey                   *[]byte
```
danielakhterov, line 26: Should be `*PublicKey`

transaction_record.go
```diff
@@ -77,6 +78,11 @@ func _TransactionRecordFromProtobuf(pb *services.TransactionRecord) TransactionR
 		tokenAssociation = append(tokenAssociation, tokenAssociationFromProtobuf(association))
 	}
 
+	var alias *[]byte
+	if len(pb.Alias) != 0 {
```
danielakhterov, line 82: Is this the correct comparison? I assume `Alias` can be `nil`

danielakhterov (changes requested):

account_id.go
andrix10: no this should be comparing string to string, as thats what String() returns

transaction_record.go
andrix10: No in go arrays are not nil just empty, and they don't have it as a pointer in protobuf

id.go
andrix10: I need it here for AccountID, others ignore the alias

danielakhterov (changes requested):

account_id.go
```diff
@@ -25,15 +26,26 @@ type _AccountIDs struct {
...
+		}, nil
+	}
+
 	return AccountID{
 		Shard:    uint64(shard),
 		Realm:    uint64(realm),
 		Account:  uint64(num),
+		AliasKey: nil,
 		checksum: checksum,
 	}, nil
```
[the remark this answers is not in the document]
andrix10: well I did that as I need to know if there is alias or not, if it returns 0 it doesn't really guarantee that there is an alias present

account_create_transaction_e2e_test.go
```diff
@@ -52,8 +52,8 @@ func TestIntegrationAccountCreateTransactionCanExecute(t *testing.T) {
 	_, err = resp.GetReceipt(env.Client)
 	require.NoError(t, err)
 
-	err = CloseIntegrationTestEnv(env, nil)
-	require.NoError(t, err)
+	//err = CloseIntegrationTestEnv(env, nil)
```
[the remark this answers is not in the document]
andrix10: I was trying to create a previewnet account with this, then realized I have to do it in CreateIntegrationTestEnv, and forgot to uncommment this

Issue #329, opened 2021-11-22
`SignedTransaction`s should be created on demand to prevent `TRANSACTION_EXPIRED` errors


danielakhterov:
### Problem

If someone submits a transaction and the nodes continuously return `BUSY` then eventually the transaction will fail with `TRANSACTION_EXPIRED` instead of recreating the transaction ID.

### Solution

Create `SignedTransaction` as needed instead of all at once.

### Task List

 - [ ] `Transaction`
    - [ ] `fromBytes()`  should lock the transaction ID list
    - [ ] `getTransactionHash()`  should lock the transaction ID list
    - [ ] `getTransactionHashPerNode()`  should lock the transaction ID list
    - [ ] `getTransactionId()`  should lock the transaction ID list
    - [ ] `getSignatures()`  should lock the transaction ID list
    - [ ] `addSignatures()`  should lock the transaction ID list
    - [ ] `setTransactionId()`  should lock the transaction ID list
    - [ ] `toBytes()`  should lock the transaction ID list
    - [ ] should generate a new transaction ID if the server returns `TRANSACTION_EXPIRED`
          **NOTE**: Use a mocking test for this.
    - [ ] `getRegenerateTransactionId()`
    - [ ] `setRegenerateTransactionId()`
    - [ ] `regenerateTransactionId` should default to `true`
          Should overwrite this field with `Client.regenerateTransactionId` when
          `Transaction.regenerateTransactionId` is the default value; similar to
          `maxTransactionFee` logic
- [ ] `Client`
    - [ ] `getDefaultRegenerateTransactionId()`
    - [ ] `setDefaultRegenerateTransactionId()`

andrix10:
This is now fixed

PR #358, opened 2021-12-28
feat: create signed transaction only as neccessary

andrix10:

Closes: #329


danielakhterov:
Can you add a mock test that shows us building a transaction ID on demand. Also, how does this affect the chunked transactions such as `FileAppendTransaction` and `TopicMessageSubmitTransaction`?

andrix10:
> Can you add a mock test that shows us building a transaction ID on demand. Also, how does this affect the chunked transactions such as `FileAppendTransaction` and `TopicMessageSubmitTransaction`?
It works in both, at least on testnet.

PR #390, opened 2022-02-03
HIP-21

andrix10:
closes: #352


address_book_query.go
```diff
@@ -0,0 +1,182 @@
...
+	"time"
+
+	"github.com/hashgraph/hedera-protobufs-go/mirror"
+	"google.golang.org/grpc/status"
+)
+
+//var rstStream = regexp.MustCompile("(?i)\\brst[^0-9a-zA-Z]stream\\b")
+
+type AddressBookQuery struct {
+	errorHandler      func(stat status.Status)
```
danielakhterov, line 17: I'm not sure if these callbacks are really warranted since this is very different from `TopicMessageQuery` since it won't exist as a background task, and instead we'll listen to the entire stream before returning.

address_book_query.go
```diff
@@ -0,0 +1,182 @@
...
+		Limit: query.limit,
+	}
+	if query.fileID != nil {
+		body.FileId = query.fileID._ToProtobuf()
+	}
+
+	return body
+}
+
+func (query *AddressBookQuery) Subscribe(client *Client) (NodeAddressBook, error) {
```
danielakhterov, line 104: I think this query should be as similar to our regular queries as we can get; meaning we should have an `Execute` method.

node_address.go
```diff
@@ -4,7 +4,7 @@ import (
 	"github.com/hashgraph/hedera-protobufs-go/services"
 )
 
-type _NodeAddress struct {
+type NodeAddress struct {
```
danielakhterov, line 7: I'm very conflicted about this change as I hacked `NodeAddress` together real quick and didn't consider it for a public API meaning we might be able to design a better `NodeAddress` if we want it to be publicly available. For starters we should make all the fields public.

danielakhterov (changes requested):
Not entirely sure where we go from here, I think we need to design a references PR first.

address_book_query.go
andrix10: Yep, still trying to figure all of this out

PR #391, opened 2022-02-03
refactor: TransferTransaction

andrix10:

### DIFF
--- a/contract_function_result.go
+++ b/contract_function_result.go
@@ -138,7 +138,6 @@ func _ContractFunctionResultFromProtobuf(pb *services.ContractFunctionResult) Co
 		Bloom:                pb.Bloom,
 		GasUsed:              pb.GasUsed,
 		LogInfo:              infos,
-		CreatedContractIDs:   createdContractIDs,
 		ContractStateChanges: csc,
 		EvmAddress:           evm,
 	}
@@ -157,12 +156,6 @@ func (result ContractFunctionResult) _ToProtobuf() *services.ContractFunctionRes
 		infos[i] = info._ToProtobuf()
 	}
 
-	contractIDs := make([]*services.ContractID, len(result.CreatedContractIDs))
-
-	for i, contractID := range result.CreatedContractIDs {
-		contractIDs[i] = contractID._ToProtobuf()
-	}
-
 	stateChanges := make([]*services.ContractStateChange, 0)
 	for _, change := range result.ContractStateChanges {
 		stateChanges = append(stateChanges, change._ToProtobuf())
@@ -175,7 +168,6 @@ func (result ContractFunctionResult) _ToProtobuf() *services.ContractFunctionRes
 		Bloom:              result.Bloom,
 		GasUsed:            result.GasUsed,
 		LogInfo:            infos,
-		CreatedContractIDs: contractIDs,
 		StateChanges:       stateChanges,
 		EvmAddress:         &wrapperspb.BytesValue{Value: result.EvmAddress.EvmAddress},
 	}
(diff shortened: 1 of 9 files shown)


contract_function_result.go
```diff
@@ -108,14 +108,6 @@ func _ContractFunctionResultFromProtobuf(pb *services.ContractFunctionResult) Co
 		infos[i] = _ContractLogInfoFromProtobuf(info)
 	}
 
-	createdContractIDs := make([]ContractID, 0)
```
danielakhterov, line 111: Not sure why this is being removed.

hbar_transfer.go
```diff
@@ -0,0 +1,74 @@
+package hedera
+
+import "github.com/hashgraph/hedera-protobufs-go/services"
+
+type HbarTransfer struct {
```
danielakhterov, line 5: In the Java and JS SDK this type is private.

hbar_transfer.go
```diff
@@ -0,0 +1,74 @@
...
+
+func (transfers *_HbarTransfers) Len() int {
+	return len(transfers.transfers)
+}
+func (transfers *_HbarTransfers) Swap(i, j int) {
+	transfers.transfers[i], transfers.transfers[j] = transfers.transfers[j], transfers.transfers[i]
+}
+
+func (transfers *_HbarTransfers) Less(i, j int) bool {
+	if transfers.transfers[i].AccountID.Shard < transfers.transfers[j].AccountID.Shard { //nolint
```
danielakhterov, line 49: Since the comparison of `AccountID` has been done several times here (and I think more times elsewhere) we should really just create a `func (a *AccountID) _Compare(b *AccountID) int { ... }`

token_transfer_private.go
```diff
@@ -0,0 +1,39 @@
+package hedera
```
danielakhterov, line 1: Not a big fan of `private` in the name of the file.

token_transfer_private.go
```diff
@@ -0,0 +1,39 @@
...
+		decimals = &temp
+	}
+
+	return &_TokenTransfer{
+		Transfers:        _HbarTransferFromProtobuf(pb.Transfers),
+		ExpectedDecimals: decimals,
+	}
+}
+
+func (transfer *_TokenTransfer) _ToProtobuf() []*services.AccountAmount {
```
danielakhterov, line 29: Hmm, I know I did the same thing in Java because I really would like to just keep `[to|from]Protobuf()` names, but perhaps we should just do `[to|from]ProtobufMany()` methods instead?

transfer_transaction.go
```diff
@@ -36,99 +30,98 @@ func NewTransferTransaction() *TransferTransaction {
...
 		tokenTransfers: tokenTransfers,
 		nftTransfers:   nftTransfers,
 	}
 }
 
-func (transaction *TransferTransaction) SetTokenTransferApproval(tokenID TokenID, accountID AccountID, approval bool) *TransferTransaction {
-	transaction.tokenApprovals[tokenID][accountID] = approval
+func (transaction *TransferTransaction) SetTokenTransferApproval(tokenID TokenID, accountID AccountID, approval bool) *TransferTransaction { //nolint
+	for token, tokenTransfer := range transaction.tokenTransfers {
+		if token.String() == tokenID.String() {
```
danielakhterov, line 63: We should really consider a private compare function for all ID types.

danielakhterov (changes requested):

contract_function_result.go
andrix10: Because is was deprecated in protobuf, and lint was complaining that I am still using deprecated variable.

hbar_transfer.go
andrix10: We actually had a `Comapre()` for AccountID and TokenID, no idea why I didn't use it before.

token_transfer_private.go
andrix10: What would be best for the name, I changed it to internal for now.

token_transfer_private.go
andrix10: I think this should be fine for now, we can just add `[to|from]ProtobufSingle()` if we need to.

PR #393, opened 2022-02-04
feat: added owner support for Approvals/Allowances, update response c…

andrix10:
closes: #392

### DIFF
--- a/contract_function_result.go
+++ b/contract_function_result.go
@@ -108,14 +108,6 @@ func _ContractFunctionResultFromProtobuf(pb *services.ContractFunctionResult) Co
 		infos[i] = _ContractLogInfoFromProtobuf(info)
 	}
 
-	createdContractIDs := make([]ContractID, 0)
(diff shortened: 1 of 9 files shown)


contract_function_result.go
```diff
@@ -108,14 +108,6 @@ func _ContractFunctionResultFromProtobuf(pb *services.ContractFunctionResult) Co
 		infos[i] = _ContractLogInfoFromProtobuf(info)
 	}
 
-	createdContractIDs := make([]ContractID, 0)
```
danielakhterov, line 111: A bit confused as to why this is being removed.

danielakhterov (changes requested):
Other than the comment everything looks good.

contract_function_result.go
andrix10: Because lint was complaining about using deprecated variable.

danielakhterov (approved):

PR #398, opened 2022-02-15
feat: added ContractCreateFlow

andrix10:

Closes: #364


danielakhterov:
@andrix10 make sure to add documentation to the new methods and types. Also for your test, can you use bytecode which is larger than 2k so it forces the use of `FileAppendTransaction`?

andrix10:
> @andrix10 make sure to add documentation to the new methods and types. Also for your test, can you use bytecode which is larger than 2k so it forces the use of `FileAppendTransaction`?

It already uses `FileAppendTransaction` in the test, that bytecode is big enough.

PR #399, opened 2022-02-16
feat: added an option to get and set paymentTransactionID in queries

andrix10:

closes: #337

### DIFF
--- a/topic_info_query.go
+++ b/topic_info_query.go
@@ -160,7 +160,9 @@ func (query *TopicInfoQuery) Execute(client *Client) (TopicInfo, error) {
 		return TopicInfo{}, err
 	}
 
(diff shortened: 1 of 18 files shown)


topic_info_query.go
```diff
@@ -158,7 +158,11 @@ func (query *TopicInfoQuery) Execute(client *Client) (TopicInfo, error) {
 		return TopicInfo{}, err
 	}
 
-	query.paymentTransactionID = TransactionIDGenerate(client.operator.accountID)
+	if !query.lockedTransactionID {
```
danielakhterov, line 161: Heh, you doubled the info on accident.

danielakhterov (changes requested):
Can you add a test that confirms that `GetTransactionID()` returns the same transaction ID as in the payment transaction?

Issue #394, opened 2022-02-12
TLS issue with mirror node


steven-sheehy:
### Description

See https://github.com/hashgraph/hedera-sdk-java/issues/855

### Steps to reproduce

See https://github.com/hashgraph/hedera-sdk-java/issues/855

### Hedera network

mainnet

### Version

latest

PR #405, opened 2022-02-18
fix: fixed TLS, implemented auto addressbook retrieval for tls, added …

andrix10:
…tls mirror node test

closes: #394


retrieve_address_book.go
```diff
@@ -0,0 +1,67 @@
...
+	client.SetMirrorNetwork([]string{"mainnet-public.mirrornode.hedera.com:5600"})
+	ledger, err := LedgerIDFromNetworkName(NetworkNameMainnet)
+	if err != nil {
+		return &NodeAddressBook{}, err
+	}
+	client.SetLedgerID(*ledger)
+
+	result, err := NewAddressBookQuery().
+		SetFileID(FileID{0, 0, 101, nil}).
+		SetLimit(14).
```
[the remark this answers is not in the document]
steven-sheehy: Here and any for any other `NewAddressBookQuery`

retrieve_address_book.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	return &result, nil
+}
+
+func GetTestnetAddressBook() (*NodeAddressBook, error) {
+	net := make(map[string]AccountID)
+	client := ClientForNetwork(net)
+
+	client.SetMirrorNetwork([]string{"hcs.testnet.mirrornode.hedera.com:5600"})
```
steven-sheehy, line 51: This repeats the mirror node URLs from `client.go`. You should just use `ClientForPreviewnet`, etc.

retrieve_address_book.go
```diff
@@ -0,0 +1,67 @@
+package hedera
+
+func GetMainnetAddressBook() (*NodeAddressBook, error) {
+	net := make(map[string]AccountID)
+	client := ClientForNetwork(net)
+
+	client.SetMirrorNetwork([]string{"mainnet-public.mirrornode.hedera.com:5600"})
```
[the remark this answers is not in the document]
andrix10: It seems we have it as 5600 in client too, and even for all sdks. Need to fix that.

network.go
```diff
@@ -72,9 +72,17 @@ func (network *_Network) _GetLedgerID() *LedgerID {
...
-
-	if network._ManagedNetwork.transportSecurity {
-		network.addressBook = _ReadAddressBookResource("addressbook/" + id.String() + ".pb")
+	var err error
+
+	if network._ManagedNetwork.transportSecurity && network.ledgerID != nil {
+		network.addressBook, err = _GetAddressBookFromMirrorNode(*network.ledgerID)
+		if err != nil {
+			var err2 error
+			network.addressBook, err2 = _ReadAddressBookResource("addressbook/" + id.String() + ".pb")
```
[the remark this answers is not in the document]
xin-hedera: did you test it? it'll panic.

client.go
```diff
@@ -188,7 +188,7 @@ var previewnetNodes = map[string]AccountID{
 	"20.150.136.89:50211":           {Account: 9},
 }
 
-var mainnetMirror = []string{"hcs.mainnet.mirrornode.hedera.com:5600"}
+var mainnetMirror = []string{"mainnet-public.mirrornode.hedera.com:443"}
```
steven-sheehy, line 191: This is a breaking change to customers using the old configuration. Mainnet public is still rate limited.  We shouldn't do this change without discussing it with leadership and rolling it out for all. Let's undo for now.

xin-hedera (changes requested):

network.go
```diff
@@ -97,15 +117,15 @@ func (network *_Network) _SetNetworkName(net NetworkName) {
 	network._SetLedgerID(*ledger)
 }
 
-func _ReadAddressBookResource(ad string) map[AccountID]NodeAddress {
+func _ReadAddressBookResource(ad string) (map[AccountID]NodeAddress, error) {
 	f, err := ioutil.ReadFile(ad)
 	if err != nil {
-		panic(err)
+		return nil, err
```
xin-hedera, line 123: the change avoids panic but doesn't really fix the bug. the function only works when there happens to be a `addressbook/xxxnet.pb` file on disk.

I'm not sure how the tests can pass before. most likely, the test application / case is run in the same directory of the sdk source, that's why it can read the file.

xin-hedera (changes requested):

network.go
danielakhterov: Fixed this the same we fixed it in the JS SDK which is to include the hex for each address book within the SDK. This isn't a great fix, but its the only fix I could come up with because Go doesn't have a "resources/" directory like Java does for published packages :/

danielakhterov:
I'm taking over for this PR because @andrix10 has the rest of the week off. I undid his changes to query the address book on demand since I think that is a separate task altogether although it does make sense, just should be separate task. I added the same managed node address test we have in Java and JS into here to confirm our `ToSecure` and `_ToInsecure` functions work as intended as well. I did add the same test for `TopicMessageQuery` that I did for JS using the public mirror node, but it might be done right now? I'm not completely sure, but the JS SDK test is also fail which I'm fairly sure was passing last time I tried.

tls_e2e_test.go
```diff
@@ -4,11 +4,20 @@
 package hedera
 
 import (
+	"encoding/hex"
 	"testing"
 
 	"github.com/stretchr/testify/require"
 )
 
+func TestFetchCurrentBalance(t *testing.T) {
```
danielakhterov, line 13: Woops, this needs to be removed.

managed_node_address.go
```diff
@@ -49,9 +52,9 @@ func (address *_ManagedNodeAddress) _ToInsecure() *_ManagedNodeAddress {
 }
 
 func (address *_ManagedNodeAddress) _ToSecure() *_ManagedNodeAddress {
-	var port uint32
+	port := address.port
```
danielakhterov, line 55: Make sure to do the same for `_ToInSecure()`

danielakhterov (commented):
Just leaving some comments for myself.

topic_message_query_e2e_test.go
```diff
@@ -323,3 +323,31 @@ func TestIntegrationTopicMessageQueryNoStartTime(t *testing.T) {
 	err = CloseIntegrationTestEnv(env, nil)
 	require.NoError(t, err)
 }
+
+func TestIntegrationTopicMessageQueryCanExecuteWithTls(t *testing.T) {
+	client := ClientForNetwork(map[string]AccountID{})
+	client.SetMirrorNetwork([]string{"mainnet-public.mirrornode.hedera.com:443"})
```
xin-hedera, line 329: no `client.SetTransportSecurity(true)`?

xin-hedera: nvm. just saw ManagedNode sets TLS accordingly depending on the port number.

For the test case failure, you can keep the topic id and set limit to 10. there is dynamic performance issue.

danielakhterov: TYSM for the limit 10 recommendation. That did the trick, but definitely a weird issue for limit of 1 to not work. Anyways I force set `SetTransportSecurity(true)` just in case.

danielakhterov:
Since this PR does seem to solve the issue with TLS with an accompanying test I will go ahead and merge. Feel free to still review though as we can create another PR to resolve any comments made on this PR.

Issue #417, opened 2022-03-01
Checksums are broken WRT LedgerIds 


SimiHunjan:
### Problem

Reference issue https://github.com/hashgraph/hedera-sdk-java/issues/825

### Solution

Reference issue https://github.com/hashgraph/hedera-sdk-java/issues/825

PR #419, opened 2022-03-02
fix: checksums are now fixed for similar ledgerIDs, ex: 0x01 and 0x10

andrix10:

closes: #417

### DIFF
--- a/account_id_unit_test.go
+++ b/account_id_unit_test.go
@@ -35,3 +35,12 @@ func TestUnitAccountIDFromStringAlias(t *testing.T) {
 
 	assert.Equal(t, id.String(), id2.String())
 }
+
+func TestUnitChecksum(t *testing.T) {
+	ad1, err := _ChecksumParseAddress("01", "0.0.3")
+	require.NoError(t, err)
+	ad2, err := _ChecksumParseAddress("10", "0.0.3")
+	require.NoError(t, err)
+
+	require.NotEqual(t, ad1.correctChecksum, ad2.correctChecksum)
+}
--- a/entity_id.go
+++ b/entity_id.go
@@ -1,6 +1,7 @@
 package hedera
 
 import (
+	"encoding/hex"
 	"fmt"
 	"regexp"
 	"strconv"
@@ -55,7 +56,12 @@ func _ChecksumParseAddress(ledgerID string, address string) (_ParseAddressResult
 
 	ad := fmt.Sprintf("%s.%s.%s", matchArray[1], matchArray[2], matchArray[3])
 
-	checksum := _CheckChecksum(ledgerID, ad)
+	l, err := LedgerIDFromString(ledgerID)
+	if err != nil {
+		return _ParseAddressResult{status: 0}, err
+	}
+
+	checksum := _CheckChecksum(l.ToBytes(), ad)
 
 	var status int
 	switch m := matchArray[4]; {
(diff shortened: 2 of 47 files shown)


entity_id.go
```diff
@@ -94,17 +100,14 @@ func _CheckChecksum(ledgerID string, address string) string {
 	asciiA := []rune("a")[0]
 	w := 31
 
-	id := ledgerID + "000000000000"
+	zeros, _ := hex.DecodeString("000000000000")
+
+	id := append(ledgerID, zeros...)
```
danielakhterov, line 105:
```suggestion
	id := append(ledgerID, []byte{0, 0, 0, 0, 0, 0})
```

account_id_unit_test.go
```diff
@@ -35,3 +35,12 @@ func TestUnitAccountIDFromStringAlias(t *testing.T) {
...
 	assert.Equal(t, id.String(), id2.String())
 }
+
+func TestUnitChecksum(t *testing.T) {
+	ad1, err := _ChecksumParseAddress("1", "0.0.3")
+	require.NoError(t, err)
+	ad2, err := _ChecksumParseAddress("10", "0.0.3")
+	require.NoError(t, err)
+
+	require.NotEqual(t, ad1.correctChecksum, ad2.correctChecksum)
```
danielakhterov, line 45: Is this actually the case? I don't think we want 1 and 10 to equal the same checksum anymore, we actually want it to be different now.

danielakhterov (changes requested):

entity_id.go
```diff
@@ -94,17 +100,15 @@ func _CheckChecksum(ledgerID string, address string) string {
 	asciiA := []rune("a")[0]
 	w := 31
 
-	id := ledgerID + "000000000000"
+	zeros, _ := hex.DecodeString("000000000000")
```
danielakhterov, line 103: There is no point in decoding a string of 0's. If you know the value you can just hard code it. In fact you might be able to create a `const` for this.

danielakhterov (changes requested):

entity_id.go
```diff
@@ -94,17 +100,14 @@ func _CheckChecksum(ledgerID string, address string) string {
...
+
+	id := append(ledgerID, zeros...)
 	h := make([]int64, 0)
 
-	for i := 0; i < len(id); i += 2 {
-		processed, _ := strconv.ParseInt(id[i:i+2], 16, 64)
+	for i := 0; i < len(id); i += 1 {
+		processed, _ := strconv.ParseInt(hex.EncodeToString(id[i:i+1]), 16, 64)
```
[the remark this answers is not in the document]
andrix10: I am confused what yoiu mean, im converting it from base 16, and bit size 64 to int64, so far this is the only way I know how to do that.

account_id_unit_test.go
andrix10: Ye this is the test for that.

entity_id.go
andrix10: Wait i didn't see this comment yesterday some reason, but yes its fixed.

ledger_id.go
```diff
@@ -26,7 +26,11 @@ func LedgerIDFromString(id string) (*LedgerID, error) {
 			_LedgerIDBytes: []byte{2},
 		}, nil
 	default:
-		temp, err := hex.DecodeString(id)
+		id2 := id
+		if len(id)%2 != 0 {
+			id2 = "0" + id
```
danielakhterov, line 31: Don't do this, let it die in the `hex.DecodeString()` instead.

danielakhterov (changes requested):

Issue #414, opened 2022-02-24
Topic subscription does not receive messages unless StartTime has been defined. 


vectorman1:
### Description

I was following the [documentation example](https://docs.hedera.com/guides/getting-started/try-examples/submit-your-first-message) on creating a topic, subscribing, and sending some messages. 

While testing, I noticed that none of my messages were being received. Upon investigation, it was found that this was caused by the fact that I had not set a Start Time for my subscription. 

When this is set, regardless if it's a value of `time.Unix(0, 0)` or `time.Now()`, messages come through successfully. 

My expectation is that regardless of whether Start Time is set at all, or has a default value, this should have no effect.

### Steps to reproduce

1. Create a topic
2. Subscribe to it
3. Send some messages

### Hedera network

testnet

### Version

v0.2.9

### Operating system

macOS

PR #421, opened 2022-03-03
feat: updated TopicMessageQuery to have a default start time

andrix10:

closes: #414

### DIFF
--- a/topic_message_query.go
+++ b/topic_message_query.go
@@ -2,6 +2,7 @@ package hedera
 
 import (
 	"context"
+	"github.com/hashgraph/hedera-protobufs-go/services"
 	"io"
 	"math"
 	"regexp"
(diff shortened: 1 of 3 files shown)


topic_message_query.go
```diff
@@ -27,11 +27,13 @@ type TopicMessageQuery struct {
 }
 
 func NewTopicMessageQuery() *TopicMessageQuery {
+	timeNow := time.Now()
 	return &TopicMessageQuery{
 		maxAttempts:       maxAttempts,
 		errorHandler:      _DefaultErrorHandler,
 		retryHandler:      _DefaultRetryHandler,
 		completionHandler: _DefaultCompletionHandler,
+		startTime:         &timeNow,
```
steven-sheehy, line 36: This shouldn't be necessary. The server will default startTime to now if not sent. The problem must lie elsewhere. Do you have tests for this?

topic_message_query.go
```diff
@@ -27,11 +27,13 @@ type TopicMessageQuery struct {
 }
```
steven-sheehy, line 27: There's a bug in this file. It should be assigning body.ConsensusEndTime:
```go
	if query.endTime != nil {
		body.ConsensusStartTime = _TimeToProtobuf(*query.endTime)
	}
```

steven-sheehy (changes requested):

topic_message_query.go
andrix10: Thanks for catching this

topic_message_query.go
andrix10: Yed we had an e2e test, but it was actually erroring in the end with timeout. So far I am not sure at all what causes it, start time is nil whole time if not set. But if i just set it to empty it works.

steven-sheehy (approved):

PR #461, opened 2022-04-12
Feature/account details query

andrix10:


account_details.go
```diff
@@ -0,0 +1,207 @@
+package hedera
```
danielakhterov, line 1: Needs to be removed.

danielakhterov (changes requested):

Issue #463, opened 2022-04-12
Regression Freeze() doesn't set body.NodeAccountID which breaks external signing


xin-hedera:
https://github.com/hashgraph/hedera-sdk-go/blob/b91bd3b3de9da0feb769b23a50ed343fe93a3a64/transaction.go#L473-L482

Issue #462, opened 2022-04-12
Regression *Transaction.GetTransactionID() panics if transaction id is not set


xin-hedera:
### Description

error message:

```
panic: runtime error: index out of range [0] with length 0 [recovered]
```

stack trace:

```
panic({0x498ad80, 0xc000039e00})
	/usr/local/Cellar/go/1.18/libexec/src/runtime/panic.go:838 +0x207
github.com/hashgraph/hedera-sdk-go/v2.(*_LockableSlice)._Get(...)
	/Users/xin/work/go/pkg/mod/github.com/hashgraph/hedera-sdk-go/v2@v2.12.0/lockable_slice.go:46
github.com/hashgraph/hedera-sdk-go/v2.(*_LockableSlice)._GetCurrent(...)
	/Users/xin/work/go/pkg/mod/github.com/hashgraph/hedera-sdk-go/v2@v2.12.0/lockable_slice.go:74
github.com/hashgraph/hedera-sdk-go/v2.(*Transaction).GetTransactionID(...)
	/Users/xin/work/go/pkg/mod/github.com/hashgraph/hedera-sdk-go/v2@v2.12.0/transaction.go:776
github.com/hashgraph/hedera-sdk-go/v2.(*AccountCreateTransaction).GetTransactionID(...)
	/Users/xin/work/go/pkg/mod/github.com/hashgraph/hedera-sdk-go/v2@v2.12.0/account_create_transaction.go:435
```

### Steps to reproduce

1. create a new transaction
2. call transaction.GetTransactionID()

in v2.11.0 GetTransactionID will return the default value instead of panic

### Hedera network

other

### Version

v2.12.0

PR #464, opened 2022-04-13
fix: Transaction.GetTransactionID() panic if TransactionID not set, T…

andrix10:
…ransactionIDs and NodeAccountIDs not locking during Transaction.FromBytes()

Closes: #462
Closes: #463 
Closes: #454

### DIFF
--- a/lockable_slice.go
+++ b/lockable_slice.go
@@ -98,7 +98,6 @@ func (this *_LockableSlice) _Advance() int { //nolint
 	index := this.index
 	if len(this.slice) != 0 {
 		this.index = (this.index + 1) % len(this.slice)
-		index = this.index
 	}
 	return index
 }
--- a/lockable_slice.go
+++ b/lockable_slice.go
@@ -1,5 +1,25 @@
 package hedera
 
+/*-
+ *
+ * Hedera Go SDK
+ *
+ * Copyright (C) 2020 - 2022 Hedera Hashgraph, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
 type _LockableSlice struct {
 	slice  []interface{}
 	locked bool
--- a/lockable_slice.go
+++ b/lockable_slice.go
@@ -78,6 +78,7 @@ func (this *_LockableSlice) _Advance() int { //nolint
 	index := this.index
 	if len(this.slice) != 0 {
 		this.index = (this.index + 1) % len(this.slice)
+		index = this.index
 	}
 	return index
 }
--- a/multi_app_transfer_test.go
+++ b/multi_app_transfer_test.go
@@ -57,10 +57,6 @@ func TestIntegrationMultiAppTransfer(t *testing.T) {
 		panic("Did not receive `TransferTransaction` back from signed bytes")
 	}
 
-	for _, s := range signedTx.GetNodeAccountIDs() {
-		println(s.String())
-	}
-
 	response, err := signedTx.Execute(env.Client)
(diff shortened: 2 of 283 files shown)


transaction.go
```diff
@@ -458,30 +484,28 @@ func _TransactionFreezeWith(
...
-	if err != nil {
-		// This should be unreachable
-		// From the documentation this appears to only be possible if there are missing proto types
-		panic(err)
-	}
-
-	for range transaction.nodeAccountIDs.slice {
+	for _, nodeAccountID := range transaction.nodeAccountIDs.slice {
+		body.NodeAccountID = nodeAccountID.(AccountID)._ToProtobuf()
+		bodyBytes, err := protobuf.Marshal(body)
```
xin-hedera, line 503: is there a test case?

the license header should really go to its separate PR. it's difficult to review the changes with so much noise.

transaction.go
```diff
@@ -773,7 +797,12 @@ func (this *Transaction) SetTransactionValidDuration(duration time.Duration) *Tr
 }
 
 func (this *Transaction) GetTransactionID() TransactionID {
-	return this.transactionIDs._GetCurrent().(TransactionID)
+	if this.transactionIDs._Length() > 0 {
+		t := this.transactionIDs._GetCurrent().(TransactionID)
+		return t
+	}
+
+	return TransactionID{}
```
xin-hedera, line 805: is there a test case to prevent such regression?

transaction.go
```diff
@@ -458,30 +484,28 @@ func _TransactionFreezeWith(
...
 			for _, nodeAccountID := range client.network._GetNodeAccountIDsForExecute() {
 				transaction.nodeAccountIDs._Push(nodeAccountID)
 			}
-		}
 
-		if client.defaultRegenerateTransactionIDs != transaction.regenerateTransactionID {
-			transaction.regenerateTransactionID = client.defaultRegenerateTransactionIDs
+			if client.defaultRegenerateTransactionIDs != transaction.regenerateTransactionID {
+				transaction.regenerateTransactionID = client.defaultRegenerateTransactionIDs
+			}
```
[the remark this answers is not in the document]
andrix10: yes, as if a user sets node account id, the regeneration is turned off

transaction.go
andrix10: Yes license part was my bad, it was a fast one so i thought no point of creating another pr. I added the multi_app_test and [these.](https://github.com/hashgraph/hedera-sdk-go/pull/464/commits/9cb9e978952731a9bc4bfa0c8338514a29858ce0l)

transaction.go
andrix10: Just had it added to one of the new unit tests.

transaction.go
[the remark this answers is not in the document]
andrix10: Nvm, for some reason I was thinking that node id gets modified in execute for regenerate so i wanted to simplify it, but it doesn't. I will be changing it back.

lockable_slice.go
```diff
@@ -78,6 +78,7 @@ func (this *_LockableSlice) _Advance() int { //nolint
 	index := this.index
 	if len(this.slice) != 0 {
 		this.index = (this.index + 1) % len(this.slice)
+		index = this.index
```
danielakhterov, line 101: Not sure if this is what we want to do. You're free to deviate from what we have In JS, but initially what I had in mind for `Advance` was to increment index and return previous index, not return the new incremented index. This change feels like it should cause some tests to fail and a bit concerned if that wasn't the case.

multi_app_transfer_test.go
```diff
@@ -0,0 +1,70 @@
...
+
+	switch t := tx.(type) {
+	case TransferTransaction:
+		signedTx = t
+	default:
+		panic("Did not receive `TransferTransaction` back from signed bytes")
+	}
+
+	for _, s := range signedTx.GetNodeAccountIDs() {
+		println(s.String())
```
danielakhterov, line 61: Shouldn't have print statements :)

transaction.go
```diff
@@ -165,7 +165,13 @@ func TransactionFromBytes(data []byte) (interface{}, error) { // nolint
 		}
 	}
 
-	tx.transactionIDs.locked = true
+	if tx.transactionIDs._Length() > 0 {
```
danielakhterov, line 188: Maybe we should create a method `LockableList._LockIfNotEmpty()` so you don't have to repeat yourself?

lockable_slice.go
andrix10: It seems nothing actually uses the returned int.

Issue #534, opened 2022-07-13
Transfering ERC-20 tokens


silverstr89:
### Description

In solidity i have some sample of code

        require(token.balanceOf(address(msg.sender)) >= (_amount), "Not enough");
        require(token.transferFrom(address(msg.sender), hotWallet, ( _amount)), "cannot transfer");

before i called approve and it is ok

when i trying to call function which transfering ERC-20, i got error "Reverted". and problem is definetly in hedera, because in ethereum it works fine

### Steps to reproduce

create any function which contains transfering tokens

        require(token.balanceOf(address(msg.sender)) >= (_amount), "Not enough");
        require(token.transferFrom(address(msg.sender), hotWallet, ( _amount)), "cannot transfer");

call it from hedera sdk like a 

	tx, err := hed.NewContractExecuteTransaction().
		SetContractID(c.id).
		SetGas(1_000_000).
		SetFunction("call", params).
		Execute(c.client)

### Hedera network

testnet

### Version

2.16.0

danielakhterov:
Can you add the snippet for your `params` and also what the function signature for `call` is.

silverstr89:
```
params, err := hed.NewContractFunctionParameters().
		AddAddress(to.ToSolidityAddress())
	if err != nil {
		return err
	}
	params = params.
		AddUint256(id.Bytes()).
		AddUint256(amount.Bytes())
```


```
function mint(
        address _to,
        uint256 _id,
        uint256 _amount
    ) public 
```

danielakhterov:
In the above code you're setting the function name to be `call`, but in the solidity code the function is called `mint`. Perhaps changing that would fix the issue?

silverstr89:
no, it is mint, i copied wrong

silverstr89:
I'm trying to write function which will be transfer erc-20 tokens in hedera hashgraph.

```
contract Coins is ERC20 {
    constructor(string memory name_, string memory symbol_)  ERC20(name_, symbol_) {
        _mint(msg.sender, 1000000000000000000000);
    }
}
```

very simple solidity contract

then i deploy it in hedera

```
tx, err := hedera.NewContractCreateFlow().
        SetBytecode(bytecode).
        SetGas(1_000_000).
        SetConstructorParameters(
            hedera.NewContractFunctionParameters().
                AddString("Coins").
                AddString("coin"),
        ).
        Execute(client)
```

then i've trying to transfer some erc-20 tokens to other address

```
[...]

PR #535, opened 2022-07-14
added abi support for uint256

silverstr89:
**Description**: Added abi package from go ethereum. It needs to use it for converting big.Int into uint256. See my issue for understanding problem.

**Related issue(s)**: https://github.com/hashgraph/hedera-sdk-go/issues/534

Fixes #

**Notes for reviewer**:

**Checklist**

- [X] Documented (Code comments, README, etc.)

### DIFF
--- a/abi/Readme.md
+++ b/abi/Readme.md
@@ -1,0 +1,13 @@
+# ABI package for using uint256
+
+> Regarding this [Issue]: hedera-sdk do not support uint256 conversion from big.Int.
+> Solution is usage of go-ethereum abi library.
+
+[Issue]: https://github.com/hashgraph/hedera-sdk-go/issues/534
+
+
+## Example
+
+```
+AddUint256(abi.U256(big.NewInt(some number)))
+```
--- a/abi/abi.go
+++ b/abi/abi.go
@@ -1,0 +1,146 @@
+// Copyright 2015 The go-ethereum Authors
+// This file is part of the go-ethereum library.
+//
+// The go-ethereum library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// The go-ethereum library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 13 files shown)


danielakhterov:
@silverstr89 
What's the purpose of adding this package to this repo? If you need access to ethereum ABI package shouldn't you simply use etheruem's package directly?

Also, reviewing your issue, I'm not sure what the SDK is doing wrong.

silverstr89:
> @silverstr89 What's the purpose of adding this package to this repo? If you need access to ethereum ABI package shouldn't you simply use etheruem's package directly?
> 
> Also, reviewing your issue, I'm not sure what the SDK is doing wrong.

reason is hedera sdk do not support correct conversation into uint256. and yes, it does wrong, i spent a lot of time for understanding why i cannot use big int bytes in uint256, as i could do it in go-ethereum. and if i doesn't write solution you will never helps me in my opened issue. my pull requests will helps to developers better understanding how to do conversation into uint256

silverstr89:
@danielakhterov there is no any explanation how to use uint256 in smart contracts deployed in hedera sdk. not in docs, not in logs and so on... this is very bad for developers community. and because i got likes on [stack overflow](https://stackoverflow.com/questions/72972918/transfering-erc-20-in-hedera-hashgraph)  it means devlopers of hedera sdk not explained clear this case

danielakhterov:
> @danielakhterov there is no any explanation how to use uint256 in smart contracts deployed in hedera sdk. not in docs, not in logs and so on... this is very bad for developers community. and because i got likes on [stack overflow](https://stackoverflow.com/questions/72972918/transfering-erc-20-in-hedera-hashgraph) it means devlopers of hedera sdk not explained clear this case

This is an absolutely valid point. The SDKs do a poor job of documenting how to use the larger integer variants of contracts and we should fix that. However, _imo_ adding an `abi` package from `go-ethereum` is not the way to do that. What we should do instead is improve documentation surrounding those methods and also create an example which uses `uint256` in contracts. Does this sound reasonable to you?

Created tracking issue: https://github.com/hashgraph/hedera-sdk-go/issues/537

silverstr89:
totally agree, i may delete this abi package and write some comments how to use for uin256

danielakhterov:
Closing PR, we'll prioritize the aforementioned issue next sprint.

Issue #567, opened 2022-08-30
`AliasEvmAddress` field in `AccountID` does not serialize or deserialize correctly


Sean-Tedrow-LB:
### Description

Copy/pasted from https://github.com/hashgraph/hedera-sdk-java/issues/1112

I have gotten confirmation that the `AliasEvmAddress` field in `AccountID` should serialize as the literal bytes of the `EvmAddress`, but instead it serializes as the bytes of a `Key` proto message with the `ECDSA_secp256k1` field filled with the bytes of the `EvmAddress`

### Steps to reproduce

Look at the source code

### Hedera network

other

### Version

current version

andrix10:
I think this was fixed along with alias update pr.

PR #568, opened 2022-08-31
Feature/account create alias

andrix10:
Closes: #567 #561


id.go
```diff
@@ -68,6 +71,19 @@ func _AccountIDFromString(s string) (shard int, realm int, num int, checksum *st
 		if err2 != nil {
 			return 0, 0, 0, nil, nil, nil, err
```
danielakhterov, line 72: Not part of your PR, but shouldn't we be returning `err2` here?

danielakhterov (approved):

PR #575, opened 2022-09-19
feat: adding precompile examples

andrix10:

Closes: #559


contract_helper.go
```diff
@@ -0,0 +1,205 @@
+package hedera
```
danielakhterov, line 1: This file should exist in the example, not the SDK.

danielakhterov (changes requested):

danielakhterov (approved):

Issue #565, opened 2022-08-29
client.setOperatorWith requests two signatures when one is expected


gregscullard:
### Description

I've setup an example use of client.setOperatorWith which builds a transaction with a single node set in `SetNodeAccountIds`, yet the callback for the signature is executed twice.

Have tried to set a transactionId, freeze the transaction, with `Freeze()` and `FreezeWith()` and the result is the same, two signature requests when only one is expected.

Note, this happens before .getReceipt, so it's not the receipt causing the second signature request (that doesn't ask for one).

### Steps to reproduce

See example attached (rename to `.go`)
[main.txt](https://github.com/hashgraph/hedera-sdk-go/files/9444583/main.txt)

### Hedera network

testnet

### Version

v2.17.1

### Operating system

macOS

andrix10:
Sorry I forgot to reply yesterday, the second signing is because we do a build all for transaction hash in transaction response which also signs it again.

gregscullard:
Can this be fixed somehow, trying to get the transactions signed via a signer (client.setOperatorWith(...)) which results in the external signer being prompted to sign the transaction twice.
The signer in this instance is a hardware wallet... so the end user is asked to sign twice which doesn't make sense.

andrix10:
Alright, that does seem bad, will discuss with Daniel what to do about this.

danielakhterov:
@andrix10 I don't remember exactly how the Go SDK functions, but I'm guessing we build transactions on demand, but don't cache them for later in case we get the same node account ID.

PR #577, opened 2022-09-20
fix: transaction hash is not singing anymore

andrix10:

Closes: #565

### DIFF
--- a/one_signature_e2e_test.go
+++ b/one_signature_e2e_test.go
@@ -1,0 +1,65 @@
+//go:build all || e2e
+// +build all e2e
+
+package hedera
+
+import (
+	"github.com/stretchr/testify/require"
+	"os"
+	"sync"
+	"testing"
+)
+
+type Count struct {
+	mx    *sync.Mutex
+	count int64
+}
+
+func NewCount() *Count {
+	return &Count{mx: new(sync.Mutex), count: 0}
+}
+
+func (c *Count) Incr() {
+	c.mx.Lock()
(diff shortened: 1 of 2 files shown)


transaction.go
```diff
@@ -751,6 +750,10 @@ func (this *Transaction) _BuildTransaction(index int) (*services.Transaction, er
 		return &services.Transaction{}, errors.Wrap(err, "failed to serialize transactions for building")
 	}
 
+	this.transactionCache = append(this.transactionCache, &services.Transaction{
+		SignedTransactionBytes: data,
+	})
+
 	return &services.Transaction{
 		SignedTransactionBytes: data,
 	}, nil
```
danielakhterov, line 759:
```suggestion
	transaction := &services.Transaction{
		SignedTransactionBytes: data,
	}
	
	this.transactionCache = append(this.transactionCache, transaction)
	return transaction, nil
```

danielakhterov (changes requested):
Can you add a test that confirms the transaction signer is only called once per node.

one_signature_e2e_test.go
```diff
@@ -0,0 +1,65 @@
...
+	return count
+}
+
+var fncCount = NewCount()
+
+func fnc() {
+	fncCount.Incr()
+}
+
+func TestIntegrationOneSignature(t *testing.T) {
```
danielakhterov, line 41: @andrix10 can you create a _unit_ tests for this instead?

PR #620, opened 2022-11-03
Add exempt custom fees

bamzedev:

**Description**:
This PR adds the functionality for collectors to be exempt of custom fees.

### DIFF
--- a/examples/exempt_custom_fees/main.go
+++ b/examples/exempt_custom_fees/main.go
@@ -2,27 +2,27 @@ package main
 
 import (
 	"fmt"
-	"os"
 	"github.com/hashgraph/hedera-sdk-go/v2"
+	"os"
 )
 
 func main() {
 	client, err := hedera.ClientForName(os.Getenv("HEDERA_NETWORK"))
 	if err != nil {
-		println(err.Error(), ": error creating client")
+		fmt.Println(err.Error(), ": error creating client")
 		return
 	}
 
 	id, err := hedera.AccountIDFromString(os.Getenv("OPERATOR_ID"))
 	if err != nil {
-		println(err.Error(), ": error converting string to AccountID")
+		fmt.Println(err.Error(), ": error converting string to AccountID")
 		return
 	}
 
 	// Retrieving operator key from environment variable OPERATOR_KEY
 	key, err := hedera.PrivateKeyFromString(os.Getenv("OPERATOR_KEY"))
 	if err != nil {
-		println(err.Error(), ": error converting string to PrivateKey")
+		fmt.Println(err.Error(), ": error converting string to PrivateKey")
 		return
 	}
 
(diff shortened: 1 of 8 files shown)


examples/exempt_custom_fees/main.go
```diff
@@ -166,6 +142,27 @@ func main() {
 	 *
 	 * Collector 0.0.B sends 10_000 units of the token to 0.0.A.
 	 */
+
+	// First we transfer the amount from treasury account to second account
+	treasuryTokenTransderTransaction, err := hedera.NewTransferTransaction().
```
MiroslavGatsanoga, line 147: **nitpick:** Typo: `treasuryTokenTransderTransaction` -> `treasuryTokenTransferTransaction`‏

examples/exempt_custom_fees/main.go
```diff
@@ -166,6 +142,27 @@ func main() {
...
+		AddTokenTransfer(tokenId, id, -10000).AddTokenTransfer(tokenId, secondAccountId, 10000).
+		FreezeWith(client)
+	if err != nil {
+		fmt.Println(err)
+	}
+	treasuryTokenTransderTransactionSign := treasuryTokenTransderTransaction.Sign(key)
+
+	treasuryTokenTransferSubmit, err := treasuryTokenTransderTransactionSign.Execute(client)
```
MiroslavGatsanoga, line 155: **suggestion:** We can use directly `treasuryTokenTransderTransaction.Sign(key).Execute(client)` instead of introducing `treasuryTokenTransderTransactionSign` variable.‏

examples/exempt_custom_fees/main.go
```diff
@@ -166,6 +142,27 @@ func main() {
 	 *
 	 * Collector 0.0.B sends 10_000 units of the token to 0.0.A.
 	 */
+
+	// First we transfer the amount from treasury account to second account
+	treasuryTokenTransderTransaction, err := hedera.NewTransferTransaction().
+		AddTokenTransfer(tokenId, id, -10000).AddTokenTransfer(tokenId, secondAccountId, 10000).
+		FreezeWith(client)
+	if err != nil {
+		fmt.Println(err)
```
MiroslavGatsanoga, line 151: **question:** ‏Something I noticed now is that in this example in a lot of cases we only print the error without returning. Perhaps we should we add `return`s in all those cases?

examples/exempt_custom_fees/main.go
```diff
@@ -166,6 +142,27 @@ func main() {
...
+	treasuryTokenTransferSubmit, err := treasuryTokenTransderTransactionSign.Execute(client)
+	if err != nil {
+		fmt.Println(err)
+	}
+	treasuryTransferRecipet, err := treasuryTokenTransferSubmit.GetReceipt(client)
+	if err != nil {
+		fmt.Println(err)
+	}
+	status := treasuryTransferRecipet.Status
+	fmt.Println("Sending from treasury account to the second account - 'TransferTransaction' status: ", status)
```
MiroslavGatsanoga, line 164: **suggestion:** Here ‏we can also use directly `treasuryTransferReceipt.Status` instead of introducing separate `status` variable.

MiroslavGatsanoga (approved):

Issue #601, opened 2022-10-19
Verify `setDefaultMaxQueryPayment()` was added and `setMaxQueruPayment()` was deprecated


SimiHunjan:
### Problem

The SDKs are to remain consistent with the APIs offered in each.

Java has `setMaxQueryPayment()` marked as deprecated in favor of `setDefaultMaxQueryPayment()`
https://github.com/hashgraph/hedera-sdk-java/blob/main/sdk/src/main/java/com/hedera/hashgraph/sdk/Client.java#L1144

### Solution

Verify the same is true for the Go SDK

SimiHunjan:
Please confirm this is also true for `setMaxTransactionFee()`.

In the Java SDK `setMaxTransactionFee()` fee was marked as deprecated in favor of `setDefaultMaxTransactionFee()`

https://github.com/hashgraph/hedera-sdk-java/blob/main/sdk/src/main/java/com/hedera/hashgraph/sdk/Client.java#L1073

PR #632, opened 2022-11-21
Add default max query and  transaction fee

bamzedev:

**Description**:
This PR adds the functionality to set a default query and transaction fee.
There are 3 ways to set these fees:
1. Set the fee explicitly for the current transaction/query
2. Set the default fee for all transactions/queries 
3. Set nothing and it should use some predefined fees

The priority is in the same order.

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-sdk-go/issues/601

### DIFF
--- a/account_info_query.go
+++ b/account_info_query.go
@@ -217,7 +217,7 @@ func (query *AccountInfoQuery) Execute(client *Client) (AccountInfo, error) {
 		cost = query.queryPayment
 	} else {
 		if query.maxQueryPayment.tinybar == 0 {
-			cost = client.maxQueryPayment
+			cost = client.GetDefaultMaxQueryPayment()
 		} else {
 			cost = query.maxQueryPayment
 		}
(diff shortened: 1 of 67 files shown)


account_info_query.go
```diff
@@ -217,7 +217,7 @@ func (query *AccountInfoQuery) Execute(client *Client) (AccountInfo, error) {
 		cost = query.queryPayment
 	} else {
 		if query.maxQueryPayment.tinybar == 0 {
-			cost = client.maxQueryPayment
+			cost = client.defaultMaxQueryPayment
```
MiroslavGatsanoga, line 220: **suggestion:** ‏We can use the newly introduced `client.GetDefaultMaxQueryPayment()` method instead of accessing the unexported field directly.

account_info_query_unit_test.go
```diff
@@ -75,8 +75,6 @@ func TestUnitAccountInfoQueryGet(t *testing.T) {
 	balance.GetMaxBackoff()
 	balance.GetMaxRetryCount()
 	balance.GetPaymentTransactionID()
-	balance.GetQueryPayment()
```
MiroslavGatsanoga, line 78: **question:** ‏Why remove those two?


**suggestion:** ‏I see there are number of such tests that only invoke `balance.Get` methods without any assertions, meaning those tests provide no actual value and only increase the test coverage. Perhaps in a future PR we can review those and add assertions.

MiroslavGatsanoga (approved):

PR #633, opened 2022-11-23
Expose AccountID Equality Methods

avislash:
This PR updates the AccountID object to allow users to directly call the IsZero() and Equals() methods on an AccountID object to enable direct comparison of two AccountID objects and to check whether or not an AccountID is the Zero Account ID.


avislash:
Hi @bamzedev, I added a signature to the commit so it shows up as verified. Let me know if there's anything else that needs to be done. Thanks!

bamzedev (approved):

Issue #669, opened 2023-02-10
Panic when executing multiple transactions


rokn:
### Description

I've tried executing a big chunk of transactions 10000+ in separate goroutines. This leads to a panic almost consistently, which is in different places but it always seems to be out of bounds exception

```
panic: runtime error: index out of range [2] with length 2

goroutine 34804 [running]:
github.com/hashgraph/hedera-sdk-go/v2.(*_Network)._GetNodeAccountIDsForExecute(0x140001a4a18)
	/Users/amind/go/pkg/mod/github.com/hashgraph/hedera-sdk-go/v2@v2.19.0/network.go:154 +0x1d4
github.com/hashgraph/hedera-sdk-go/v2._TransactionFreezeWith(0x140317b6500, 0x140001a4a00, 0x14023a2c150)
	/Users/amind/go/pkg/mod/github.com/hashgraph/hedera-sdk-go/v2@v2.19.0/transaction.go:517 +0x60
github.com/hashgraph/hedera-sdk-go/v2.(*ScheduleCreateTransaction).FreezeWith(0x140317b6500, 0x140001a4a00)
	/Users/amind/go/pkg/mod/github.com/hashgraph/hedera-sdk-go/v2@v2.19.0/schedule_create_transaction.go:369 +0xa8
github.com/hashgraph/hedera-sdk-go/v2.(*ScheduleCreateTransaction).Execute(0x140317b6500, 0x140001a4a00)
	/Users/amind/go/pkg/mod/github.com/hashgraph/hedera-sdk-go/v2@v2.19.0/schedule_create_transaction.go:296 +0x5c
main.main.func1(0x3e9b?)
...

panic: runtime error: index out of range [3] with length 3

goroutine 1473 [running]:
github.com/hashgraph/hedera-sdk-go/v2.(*_ManagedNetwork)._GetNode(0x140001a2a18)
	/Users/amind/go/pkg/mod/github.com/hashgraph/hedera-sdk-go/v2@v2.19.0/managed_network.go:191 +0xc0
github.com/hashgraph/hedera-sdk-go/v2.(*_Network)._GetNode(...)
	/Users/amind/go/pkg/mod/github.com/hashgraph/hedera-sdk-go/v2@v2.19.0/network.go:89
[...]

Issue #660, opened 2023-02-01
Signs more than once per node/transaction


bamzedev:
### Description

Any type of transaction will sign more than once for the same node.

### Steps to reproduce

`hedera.NewAccountCreateTransaction().SetKey(newKey).Execute(client)` observer it is signing twice.

### Hedera network

testnet

### Version

v2.20.0

### Operating system

None

Issue #659, opened 2023-02-01
Does not try other nodes when the node is offline when mutiple nodes are set


bamzedev:
### Description

Create any type of transaction, set multiple nod account ids, execute. If the first node fails, it does not try other nodes from the locked ones, but retries the first one 10 times.

### Steps to reproduce

```hedera.NewAccountCreateTransaction().SetKey(newKey).SetNodeAccountIDs([]hedera.AccountID{{Account: 4},{Account: 6}}).Execute(client)```

### Hedera network

testnet

### Version

v2.20.0

### Operating system

None

PR #670, opened 2023-02-10
Fix network issues

bamzedev:

**Description**:
This PR modifies changes the way we sign the transactions and sync between current node and current signed transaction. It also modifies the `_BuildTransaction` to check if we have such a transaction already signed, so it does not asks for a signature. 

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-sdk-go/issues/671, https://github.com/hashgraph/hedera-sdk-go/issues/669, https://github.com/hashgraph/hedera-sdk-go/issues/660, https://github.com/hashgraph/hedera-sdk-go/issues/659

### DIFF
--- a/transaction.go
+++ b/transaction.go
@@ -63,8 +63,7 @@ type Transaction struct {
 	maxBackoff              *time.Duration
 	minBackoff              *time.Duration
 	regenerateTransactionID bool
-
-	grpcDeadline *time.Duration
+	grpcDeadline            *time.Duration
 }
 
 func _NewTransaction() Transaction {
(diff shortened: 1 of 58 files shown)


transaction.go
```diff
@@ -755,6 +738,25 @@ func (this *Transaction) _BuildTransaction(index int) (*services.Transaction, er
 	if err != nil {
 		return &services.Transaction{}, errors.Wrap(err, "failed to update this ID")
 	}
+
+	// Bellow are checks whether we need to sign the transaction or we already have the same signed
+	if bytes.Equal(signedTx.BodyBytes, updatedBody) {
+		tx := this.signedTransactions._Get(index).(*services.SignedTransaction)
```
MiroslavGatsanoga, line 744: **question:** ‏Why we get the signed transaction again (I see it's done on line 764 below as well) instead of using `signedTx`?

utilities_for_test.go
```diff
@@ -77,10 +77,11 @@ func NewIntegrationTestEnv(t *testing.T) IntegrationTestEnv {
 	} else if os.Getenv("HEDERA_NETWORK") == "localhost" {
 		network := make(map[string]AccountID)
 		network["127.0.0.1:50213"] = AccountID{Account: 3}
-		network["127.0.0.1:50214"] = AccountID{Account: 4}
-		network["127.0.0.1:50215"] = AccountID{Account: 5}
+		// network["127.0.0.1:50214"] = AccountID{Account: 4}
```
MiroslavGatsanoga, line 80: **suggestion:** ‏Is this commented out on purpose or it was missed to be uncommented? If don't need the lines perhaps we should remove them instead?

MiroslavGatsanoga (approved):

PR #677, opened 2023-02-14
Update mirror node endpoints

bamzedev:

**Description**:
Just updating mirror node endpoints.

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-sdk-go/issues/665 https://github.com/hashgraph/hedera-sdk-go/issues/663

### DIFF
--- a/tls_e2e_test.go
+++ b/tls_e2e_test.go
@@ -41,7 +41,7 @@ func TestIntegrationPreviewnetTls(t *testing.T) {
 	client := ClientForNetwork(network)
 	ledger, _ := LedgerIDFromNetworkName(NetworkNamePreviewnet)
 	client.SetTransportSecurity(true)
-	client.SetMirrorNetwork([]string{"hcs.previewnet.mirrornode.hedera.com:5600"})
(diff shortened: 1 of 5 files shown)


MiroslavGatsanoga (approved):

tls_e2e_test.go
```diff
@@ -67,7 +67,7 @@ func TestIntegrationTestnetTls(t *testing.T) {
 	ledger, _ := LedgerIDFromNetworkName(NetworkNameTestnet)
 	client.SetTransportSecurity(true)
 	client.SetLedgerID(*ledger)
-	client.SetMirrorNetwork([]string{"hcs.testnet.mirrornode.hedera.com:5600"})
+	client.SetMirrorNetwork([]string{"testnet.mirrornode.hedera.com:5600"})
```
steven-sheehy, line 70: This is not a valid endpoint. Needs to be 443. Did you run this test?

Similar comment on other tests.

bamzedev: You are correct. Yes, all tests passed. This test isn't exactly test if this is valid network, it reverts to `443`. I'll revert this commit and open another PR as it would require a lot more changes since the SDK expects that there are 2 endpoints. Thanks for noticing.

PR #686, opened 2023-02-24
fix: Change hip583 example to testnet

bamzedev:
**Description**:
Just changing the HIP583 example to run by default against `Testnet` instead of `Previewnet`


Petyo-Lukanov (approved):

Petyo-Lukanov:
LGTM

PR #696, opened 2023-03-01
Add integration tests and coverage to PR's

bamzedev:
**Description**:
Test


.github/workflows/build.yml
```diff
@@ -1,59 +1,75 @@
-name: Build
+name: PR Checks
```
ochikov, line 1: Let's name it: Build & Test?

.github/workflows/build.yml
```diff
@@ -1,59 +1,75 @@
...
+            - name: Build
+              run: go build -v .
+
+            - name: Lint
+              uses: golangci/golangci-lint-action@v3
+              with:
+                  skip-cache: true
+
+    test:
+        name: Test
```
ochikov, line 42: let's name it: Unit & Integration Test

ochikov (changes requested):

.github/workflows/build.yml
bamzedev: Wasn't sure what to use tbh, because it is not just building and running tests. It does build, some go checks, lint checks and then tests.

bamzedev: Hmm, Services also have it named that way `PR Checks`.

ochikov: Ok. Let's leave it as it is.

.github/workflows/build.yml
```diff
@@ -1,59 +1,75 @@
...
-        with:
-          version: v1.50
+    build:
+        name: Build
+        runs-on: ubuntu-latest
+        steps:
+            - name: Set up Go 1.19
+              uses: actions/setup-go@v3
+              with:
+                  go-version: 1.19
```
MiroslavGatsanoga, line 21: **suggestion:** ‏Perhaps not for this PR but I think we can easily upgrade to the latest 1.20 version.

.github/workflows/build.yml
```diff
@@ -1,59 +1,75 @@
...
+
+            - name: Tests Integration
+              if: success()
+              run: go test -tags="e2e" -timeout 9999s -v -coverprofile=e2e.out -covermode=atomic
+
+            - name: Upload coverage to Codecov
+              if: success()
+              uses: codecov/codecov-action@v3
+              with:
+                files: ./unit.out,./e2e.out
```
MiroslavGatsanoga, line 72: **question:** Why use both unit & integration tests for coverage instead of only unit tests?

ochikov (approved):

PR #726, opened 2023-04-06
Fix:Der headers

bamzedev:
**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/crypto.go
+++ b/crypto.go
@@ -35,11 +35,9 @@ import (
 	"golang.org/x/crypto/pbkdf2"
 )
 
-const ed25519PrivateKeyPrefix = "302e020100300506032b657004220420"
-const ed25519PubKeyPrefix = "302a300506032b6570032100"
+const _Ed25519PrivateKeyPrefix = "302e020100300506032b657004220420"
 
-const _ECDSAPrivateKeyPrefix = "3030020100300706052b8104000a04220420"
-const _ECDSAPubKeyPrefix = "302f300706052b8104000a0324000421"
+const _ECDSAPrivateKeyPrefix = "30540201010420"
 
 type Key interface {
 	_ToProtoKey() *services.Key
@@ -279,7 +277,7 @@ func PrivateKeyFromStringDer(s string) (PrivateKey, error) {
 		return PrivateKey{
 			ecdsaPrivateKey: key,
 		}, nil
-	} else if strings.Contains(s, ed25519PrivateKeyPrefix) {
+	} else if strings.Contains(s, _Ed25519PrivateKeyPrefix) {
 		key, err := _Ed25519PrivateKeyFromString(s)
 		if err != nil {
 			return PrivateKey{}, err
--- a/crypto_unit_test.go
+++ b/crypto_unit_test.go
@@ -35,6 +35,9 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+const _Ed25519PubKeyPrefix = "302a300506032b6570032100"
+const _ECDSAPubKeyPrefix = "3036301006072a8648ce3d020106052b8104000a0322000"
+
 const testPrivateKeyStr = "302e020100300506032b657004220420db484b828e64b2d8f12ce3c0a0e93a0b8cce7af1bb8f39c97732394482538e10"
 
 const testPublicKeyStr = "302a300506032b6570032100e0c8ec2758a5879ffac226a13c0c516b799e72e35141a0dd828f94d37988a4b7"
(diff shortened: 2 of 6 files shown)


crypto_unit_test.go
```diff
@@ -35,6 +36,9 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+const _Ed25519PubKeyPrefix = "302a300506032b6570032100"
+const _ECDSAPubKeyPrefix = "3036301006072a8648ce3d020106052b8104000a0322000"
```
MiroslavGatsanoga, line 40: **question:** ‏The issue states that the prefix for ECDSA Public key should be `302d300706052b8104000a032200`, so is the issue outdated or is this value not expected?

crypto_unit_test.go
MiroslavGatsanoga, line 1: **praise:** Very thorough with adding all of the new tests, really nice!

ecdsa_public_key.go
```diff
@@ -70,24 +73,67 @@ func _ECDSAPublicKeyFromBytesDer(byt []byte) (*_ECDSAPublicKey, error) {
...
+	}
+
+	// Check if the parsed key uses ECDSA public key algorithm
+	ecdsaPublicKeyAlgorithmOID := asn1.ObjectIdentifier{1, 2, 840, 10045, 2, 1}
+	if !key.AlgorithmIdentifier.Algorithm.Equal(ecdsaPublicKeyAlgorithmOID) {
+		return nil, errors.New("public key is not an ECDSA public key")
+	}
+
+	// Check if the parsed key uses secp256k1 curve
+	secp256k1OID := asn1.ObjectIdentifier{1, 3, 132, 0, 10}
```
MiroslavGatsanoga, line 99: **thought:** ‏We can probably extract those `asn1.ObjectIdentifier` that we use at a number of different places, have them defined in a single place and just reuse them where needed.

crypto_unit_test.go
bamzedev: Yes, the ticket is wrong. What we currently have in this PR is aligned with openSSL, which we are currently thinking are following the proper standard for DER.

ecdsa_public_key.go
bamzedev: Not sure, its used in 2 places in this file, so not that much of duplication. Happy either way.

MiroslavGatsanoga (approved):

crypto.go
```diff
@@ -35,11 +35,9 @@ import (
 	"golang.org/x/crypto/pbkdf2"
 )
 
-const ed25519PrivateKeyPrefix = "302e020100300506032b657004220420"
-const ed25519PubKeyPrefix = "302a300506032b6570032100"
+const _Ed25519PrivateKeyPrefix = "302e020100300506032b657004220420"
 
-const _ECDSAPrivateKeyPrefix = "3030020100300706052b8104000a04220420"
-const _ECDSAPubKeyPrefix = "302f300706052b8104000a0324000421"
+const _ECDSAPrivateKeyPrefix = "30540201010420"
```
dikel, line 40: If the user exports his ECDSA private key using OpenSSL and doesn't specify he wants the compressed format, the prefix will be `30740201010420`. Also the `parameters` and `publicKey` fields of `ECPrivateKey` are optional so there are a lot of valid ECDSA private key prefixes.

I think we can remove the `if strings.Contains(s, _ECDSAPrivateKeyPrefix)` and just call `_ECDSAPrivateKeyFromString`. And if there is an error, try `_Ed25519PrivateKeyFromString`

ed25519_public_key.go
```diff
@@ -115,10 +119,29 @@ func (pk _Ed25519PublicKey) _BytesRaw() []byte {
 	return pk.keyData
 }
 
-// _Bytes returns the byte slice representation of the _Ed25519PublicKey.
 func (pk _Ed25519PublicKey) _BytesDer() []byte {
-	decoded, _ := hex.DecodeString(ed25519PubKeyPrefix)
-	return append(decoded, pk._BytesRaw()...)
+	type PublicKeyInfo struct {
```
dikel, line 123: Wouldn't it be better to move all asn1 structs (ECPrivateKey, AlgorithmIdentifier, PublicKeyInfo, PrivateKeyInfo) in a separate class?

dikel (changes requested):

ed25519_public_key.go
bamzedev: We have a lot of global varaibles already, dont want to bloat it even more.

dikel (approved):
LGTM

ecdsa_private_key.go
```diff
@@ -155,6 +188,43 @@ func (sk *_ECDSAPrivateKey) _PublicKey() *_ECDSAPublicKey {
...
+
+	key, err := _ECDSAPrivateKeyFromBytes(block.Bytes)
+	if err != nil {
+		return nil, err
+	}
+	return key, nil
+}
+
+func _ECDSAPrivateKeyReadPem(source io.Reader, passphrase string) (*_ECDSAPrivateKey, error) {
+	// note: Passphrases are currently not supported, but included in the function definition to avoid breaking
```
dikel, line 217: Isn't the passphrase supported? I saw that the test vector for ECDSA private key from PEM with passphrase passes

dikel (changes requested):

dikel (approved):
LGTM

PR #736, opened 2023-05-15
Chore: Add AccountAllowanceAdjustTransaction Unit Tests

bamzedev:
**Description**:
Just ramping up the test coverage.

### DIFF
--- a/account_allowance_adjust_transaction_unit_test.go
+++ b/account_allowance_adjust_transaction_unit_test.go
@@ -33,6 +33,7 @@ import (
 
 func TestUnitAccountAllowanceAdjustTransactionGet(t *testing.T) {
 	key, err := PrivateKeyGenerate()
+	require.NoError(t, err)
 	key2, err := PrivateKeyGenerate()
 	require.NoError(t, err)
 
--- a/account_allowance_adjust_transaction_unit_test.go
+++ b/account_allowance_adjust_transaction_unit_test.go
@@ -1,0 +1,206 @@
+//go:build all || unit
+// +build all unit
+
+package hedera
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+/*-
+ *
+ * Hedera Go SDK
+ *
+ * Copyright (C) 2020 - 2022 Hedera Hashgraph, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+func TestUnitAccountAllowanceAdjustTransactionGet(t *testing.T) {
+	key, err := PrivateKeyGenerate()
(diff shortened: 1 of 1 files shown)


account_allowance_adjust_transaction_unit_test.go
```diff
@@ -0,0 +1,206 @@
...
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+/*-
+ *
+ * Hedera Go SDK
+ *
+ * Copyright (C) 2020 - 2022 Hedera Hashgraph, LLC
```
MiroslavGatsanoga, line 18: **nitpick (non-blocking):** ‏We should have 2023 in the copyright line, that actually applies to most of the files in the repo, so this could probably be addressed in a separate PR for all files.

account_allowance_adjust_transaction_unit_test.go
```diff
@@ -0,0 +1,206 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+func TestUnitAccountAllowanceAdjustTransactionGet(t *testing.T) {
+	key, err := PrivateKeyGenerate()
+	key2, err := PrivateKeyGenerate()
+	require.NoError(t, err)
```
MiroslavGatsanoga, line 37: **nitpick:** ‏In case we got an error for the first key but not for the second this require will pass instead of failing. Probably we need another require right after the first key creation?

account_allowance_adjust_transaction_unit_test.go
```diff
@@ -0,0 +1,206 @@
...
+		AddTokenAllowance(tokenID, nodeAccountIDs[0], 1).
+		AddTokenNftAllowance(NftID, nodeAccountIDs[0]).
+		SetTransactionID(transactionID).SetNodeAccountIDs(nodeAccountIDs).
+		SetMaxTransactionFee(HbarFromTinybar(100)).SetRegenerateTransactionID(true).
+		SetTransactionMemo("go sdk unit test").SetTransactionValidDuration(time.Second * 120).
+		SetMaxRetry(1).SetMaxBackoff(time.Second * 120).SetMinBackoff(time.Second * 1).
+		Freeze()
+	sign, err := key2.SignTransaction(&tx.Transaction)
+	require.NoError(t, err)
+	tx.AddSignature(key.PublicKey(), sign)
```
MiroslavGatsanoga, line 54: **question:** ‏Hmm, we use the signature that is generated with `key2.SignTransaction()` but with the other key? Actually why do we need two keys, can't we simplify the code and use only one?

account_allowance_adjust_transaction_unit_test.go
```diff
@@ -0,0 +1,206 @@
...
+			SpenderAccountID: &AccountID{Account: 3},
+			OwnerAccountID:   nil,
+			TokenID:          &tokenID,
+			SerialNumbers:   []int64{},
+			AllSerials:       true,
+			DelegatingSpender: nil,
+		},
+	}
+	require.Equal(t, expectedTokenNftAllowances, tx.GetTokenNftAllowances())
+}
```
MiroslavGatsanoga, line 206: **nitpick:** ‏Missing newline at end of file.

account_allowance_adjust_transaction_unit_test.go
bamzedev: We need it for coverage.

account_allowance_adjust_transaction_unit_test.go
bamzedev: I did bring it some time ago. Not sure what was decided for that. Decided to add it so that when we move to update it, a simple search will catch this is as well.

MiroslavGatsanoga (approved):

PR #746, opened 2023-06-02
Feature/refactor logging

### LINKED ISSUE #611: Add stats for transactions
### Problem

There are times when transactions fail with `INVALID_NODE_ACCOUNT_ID`, `TRANSACTION_EXPIRED`, or other errors where transactions do not succeed. In order to debug these transaction failures the SDK needs to provide some stats.

### Solution

Log the transactions that were submitted in the last 30 minutes.

Stored locally and default is not enabled.

This should include:

- Transaction ID
- Node the transaction was submitted to
- Network response (either the network error or success response)
- The network is was submitted to (mainnet, testnet, previewnet)

### Alternatives

_No response_

### LINKED ISSUE #627: logCtx should not be a global variable
### Problem

We should not have a logger set as a global variable at initialization. Using the `init()` function in Go is generally frowned upon because there's no easy way to tell the order in which those get executed. Having a global variable with a logger makes testing difficult and causes strange side-effects when people embed the SDK inside their programs. [See Discord issue here.](https://discord.com/channels/373889138199494658/535491142280806400/1040277366527758376)
https://github.com/hashgraph/hedera-sdk-go/blob/develop/executable.go#L44

Structs that require a default logger should be able to set the logger via a method. Functions/methods that take a context, should be able to extract the logger from the `context.Context` to use.
https://github.com/rs/zerolog#pass-a-sub-logger-by-context

## Pass a logger by context

```golang
ctx := log.With().Str("component", "module").Logger().WithContext(ctx)

log.Ctx(ctx).Info().Msg("hello world")

// Output: {"component":"module","level":"info","message":"hello world"}
```

### Alternatives

_No response_

Coiling-Dragon:
**Description**:

This pull request introduces changes to address the logger behaviour and specifications. It includes the implementation of a Logger class and LogLevel mapping, along with necessary functionalities. Additionally, it logs relevant information such as transaction details, node status, network response, and more.

* In order to accommodate the architecture of the GO SDK, it was necessary to make changes to all transaction and query files.

**Related issue(s)**:

Fixes #611 
Fixes #627

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/account_balance_query.go
+++ b/account_balance_query.go
@@ -175,8 +175,8 @@ func (query *AccountBalanceQuery) GetCost(client *Client) (Hbar, error) {
 	return HbarFromTinybar(cost), nil
 }
 
-func _AccountBalanceQueryShouldRetry(logID string, _ interface{}, response interface{}) _ExecutionState {
-	return _QueryShouldRetry(logID, Status(response.(*services.Response).GetCryptogetAccountBalance().Header.NodeTransactionPrecheckCode))
+func _AccountBalanceQueryShouldRetry(_ interface{}, response interface{}) _ExecutionState {
+	return _QueryShouldRetry(Status(response.(*services.Response).GetCryptogetAccountBalance().Header.NodeTransactionPrecheckCode))
 }
 
 func _AccountBalanceQueryMapStatusError(_ interface{}, response interface{}) error {
@@ -310,5 +310,10 @@ func (query *AccountBalanceQuery) _GetLogID() string {
 // SetPaymentTransactionID assigns the payment transaction id.
 func (query *AccountBalanceQuery) SetPaymentTransactionID(transactionID TransactionID) *AccountBalanceQuery {
 	query.paymentTransactionIDs._Clear()._Push(transactionID)._SetLocked(true)
+	return query
+}
+
+func (query *AccountBalanceQuery) SetLogLevel(level LogLevel) *AccountBalanceQuery {
+	query.Query.SetLogLevel(level)
 	return query
 }
--- a/examples/logging/main.go
+++ b/examples/logging/main.go
@@ -1,0 +1,80 @@
+package main
+
+import (
+	"fmt"
+	"github.com/hashgraph/hedera-sdk-go/v2"
+	"os"
+)
+
+func main() {
+	var client *hedera.Client
+	var err error
+
+	// Retrieving network type from environment variable HEDERA_NETWORK
+	client, err = hedera.ClientForName(os.Getenv("HEDERA_NETWORK"))
+	if err != nil {
+		println(err.Error(), ": error creating client")
+		return
+	}
+
+	// The client comes with default logger.
+	// We can also set a custom logger for this client.
+	// The logger must implement the hedera.Logger interface.
+	logger := hedera.NewLogger("Hedera sdk", hedera.LoggerLevelDebug)
+	client.SetLogger(logger)
+
+	// Set the logging level fot this client, to be used as default.
+	// Individual log levels can be set for the Query or Transaction object by
+	// chaining the SetLogLevel() function on the given Transaction or Query object.
+	client.SetLogLevel(hedera.LoggerLevelTrace)
+
+	// Retrieving operator ID from environment variable OPERATOR_ID
+	operatorAccountID, err := hedera.AccountIDFromString(os.Getenv("OPERATOR_ID"))
+	if err != nil {
+		println(err.Error(), ": error converting string to AccountID")
+		return
+	}
+
+	// Retrieving operator key from environment variable OPERATOR_KEY
+	operatorKey, err := hedera.PrivateKeyFromString(os.Getenv("OPERATOR_KEY"))
+	if err != nil {
+		println(err.Error(), ": error converting string to PrivateKey")
+		return
+	}
+
+	// Setting the client operator ID and key
+	client.SetOperator(operatorAccountID, operatorKey)
+
+	// Generate new key to use with new account
+	newKey, err := hedera.GeneratePrivateKey()
+	if err != nil {
+		println(err.Error(), ": error generating PrivateKey}")
+		return
+	}
+
+	fmt.Printf("private = %v\n", newKey)
+	fmt.Printf("public = %v\n", newKey.PublicKey())
+
+	// Transaction used to show default logging functionality from client
+	_, err = hedera.NewAccountCreateTransaction().
+		SetKey(newKey.PublicKey()).
+		Execute(client)
+	if err != nil {
+		println(err.Error(), ": error executing account create transaction}")
+		return
+	}
+
+	// Disable default logging on client, to show logging functionality from transaction
+	client.SetLogLevel(hedera.LoggerLevelDisabled)
+
+	// Create account transaction used to show logging functionality
+	_, err = hedera.NewAccountCreateTransaction().
+		SetKey(newKey.PublicKey()).
+		// Set logging level for the specific transaction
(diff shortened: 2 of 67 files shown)


query.go
```diff
@@ -137,7 +137,6 @@ func (this *Query) SetMaxRetry(count int) *Query {
 }
 
 func _QueryShouldRetry(logID string, status Status) _ExecutionState {
-	logCtx.Trace().Str("requestId", logID).Str("status", status.String()).Msg("query precheck status received")
```
Coiling-Dragon, line 140: To remove `_QueryShouldRetry` - `logID` there will be changes in all transaction files ~20 files.

logger.go
```diff
@@ -0,0 +1,153 @@
...
+	Error(msg string, keysAndValues ...interface{})
+	Warn(msg string, keysAndValues ...interface{})
+	Info(msg string, keysAndValues ...interface{})
+	Debug(msg string, keysAndValues ...interface{})
+	Trace(msg string, keysAndValues ...interface{})
+}
+
+type DefaultLogger struct {
+	logger *zerolog.Logger
+	level  *LogLevel
```
rokn, line 54: Do we need the log level to be a pointer for some reason?

address_book_query.go
```diff
@@ -191,3 +192,12 @@ func (query *AddressBookQuery) Execute(client *Client) (NodeAddressBook, error)
...
 }
+
+func (query *AddressBookQuery) SetLogLevel(level LogLevel) *AddressBookQuery {
+	query.logLevel = &level
+	return query
+}
+
+func (query *AddressBookQuery) GetLogLevel() *LogLevel {
+	return query.logLevel
+}
```
rokn, line 203: This is not used anywhere because of the custom flow of this query. We should probably remove it for now

account_balance_query.go
```diff
@@ -312,3 +312,12 @@ func (query *AccountBalanceQuery) SetPaymentTransactionID(transactionID Transact
...
 }
+
+func (query *AccountBalanceQuery) SetLogLevel(level LogLevel) *AccountBalanceQuery {
+	query.Query.logLevel = &level
+	return query
+}
+
+func (query *AccountBalanceQuery) GetLogLevel() *LogLevel {
+	return query.Query.logLevel
+}
```
rokn, line 323: It would probably be better if all query and transactions use the base Query/Transaction methods for setting/getting log level.

examples/create_account/main.go
```diff
@@ -56,11 +57,15 @@ func main() {
...
+		SetTransactionMemo("go sdk example create_account/main.go")
+		// Freeze()
 		// The account is charged to extend its expiration date every this many seconds. If it doesn't
 		// have enough balance, it extends as long as possible. If it is empty when it expires, then it
 		// is deleted.
-		Execute(client)
+		// Execute(client)
+
+	transactionResponse, err := tx.Execute(client)
+
```
rokn, line 68: Is this leftover from testing?

executable.go
```diff
@@ -239,15 +211,56 @@ func _Execute( // nolint
 		}
 
 		node._DecreaseBackoff()
-		switch shouldRetry(logID, request, resp) {
+
+		// Check if the response is a querry or a transaction
```
rokn, line 215: nit: 
```suggestion
		// Check if the response is a query or a transaction
```

executable.go
```diff
@@ -239,15 +211,56 @@ func _Execute( // nolint
...
+		if isTransaction {
+			if tx.logLevel != nil {
+				txLogger = client.logger.SetLevel(*tx.logLevel)
+			}
+		} else {
+			q, _ := request.(*Query)
+			if q.logLevel != nil {
+				txLogger = client.logger.SetLevel(*q.logLevel)
+			}
+		}
```
rokn, line 227: Maybe this should happen outside of the big for loop and it could be in a utility function to increase readibility

rokn (approved):

account_balance_query.go
Coiling-Dragon: I will leave the `GetLogLevel()` method only in the base Query and Transaction structs as a method ( removed from the types of transactions and queries )

refactoring the `SetLogLevel()` to use the base methods like so:
`
...{
    transaction.Transaction.SetLogLevel(level)
    return transaction
}
`

examples/create_account/main.go
Coiling-Dragon: Yes, i will revert the changes

address_book_query.go
Coiling-Dragon: Sure, for now it is not needed.

logger.go
Coiling-Dragon: It is leftover from previous refactoring. Updating it.

address_book_query.go
Coiling-Dragon: removed

executable.go
Coiling-Dragon: refactored

executable.go
Coiling-Dragon: deleted entire comment due to refactor

executable.go
```diff
@@ -173,18 +178,17 @@ func _Execute( // nolint
...
 		node._InUse()
 
-		logCtx.Trace().Str("requestId", logID).Str("nodeAccountID", node.accountID.String()).Str("nodeIPAddress", node.address._String()).
-			Str("Request Proto:", hex.EncodeToString(marshaledRequest)).Msg("executing")
+		client.logger.Trace("executing", "requestId", logID, "nodeAccountID", node.accountID.String(), "nodeIPAddress", node.address._String(), "Request Proto", hex.EncodeToString(marshaledRequest))
 
 		if !node._IsHealthy() {
-			logCtx.Trace().Str("requestId", logID).Str("delay", node._Wait().String()).Msg("node is unhealthy, waiting before continuing")
-			_DelayForAttempt(logID, backOff.NextBackOff(), attempt)
+			client.logger.Trace("node is unhealthy, waiting before continuing", "requestId", logID, "delay", node._Wait().String())
```
rokn, line 184: I think everywhere in the execute function we should use the `txLogger` defined above

Coiling-Dragon: fair point

logger.go
```diff
@@ -0,0 +1,153 @@
...
+func (l *DefaultLogger) SetSilent(isSilent bool) {
+	if isSilent {
+		logger := l.logger.Level(zerolog.Disabled)
+		l.logger = &logger
+	} else {
+		l.SetLevel(l.level)
+	}
+}
+
+func logger_for_level(logger zerolog.Logger, level LogLevel) zerolog.Logger {
```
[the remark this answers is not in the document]
Coiling-Dragon: again fair point

examples/logging/main.go
```diff
@@ -0,0 +1,80 @@
...
+
+	// Retrieving network type from environment variable HEDERA_NETWORK
+	client, err = hedera.ClientForName(os.Getenv("HEDERA_NETWORK"))
+	if err != nil {
+		println(err.Error(), ": error creating client")
+		return
+	}
+
+	// The client comes with default logger.
+	// We can also set a custom logger for this clien.
```
bamzedev, line 21: typo. but w/e we can fix with the PR that would be adding tests.

bamzedev:
You can add this issue to the PR as well https://github.com/hashgraph/hedera-sdk-go/issues/627

bamzedev (approved):

examples/logging/main.go
Coiling-Dragon: Yea, thanks

rokn (approved):

PR #751, opened 2023-06-12
Added token transactions tests

rokn:
**Description**:
This PR adds some test for some of the token transactions and info objects.
It also adds a small improvement to the token balance map.

**Checklist**

### DIFF
--- a/token_pause_transaction_e2e_test.go
+++ b/token_pause_transaction_e2e_test.go
@@ -1,0 +1,66 @@
+//go:build all || e2e
+// +build all e2e
+
+package hedera
+
+import (
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func TestIntegrationTokenPauseTransactionCanExecute(t *testing.T) {
+	env := NewIntegrationTestEnv(t)
+
+	resp, err := NewTokenCreateTransaction().
+		SetNodeAccountIDs(env.NodeAccountIDs).
+		SetTokenName("ffff").
+		SetTokenSymbol("F").
+		SetAdminKey(env.Client.GetOperatorPublicKey()).
+		SetTreasuryAccountID(env.Client.GetOperatorAccountID()).
+		SetPauseKey(env.Client.GetOperatorPublicKey()).
+		SetFreezeDefault(false).
+		Execute(env.Client)
+	require.NoError(t, err)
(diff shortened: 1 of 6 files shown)


token_pause_transaction_e2e_test.go
```diff
@@ -0,0 +1,111 @@
...
+	info, err = NewTokenInfoQuery().
+		SetTokenID(tokenID).
+		SetNodeAccountIDs([]AccountID{resp.NodeID}).
+		Execute(env.Client)
+	require.NoError(t, err)
+
+	assert.NotNil(t, info.PauseStatus)
+	assert.True(t, *info.PauseStatus)
+
+	//Unpause token to avoid error in CloseIntegrationTestEnv
```
bamzedev, line 54: If the test fails in some of the steps above, this wont be executed. `defer()` maybe?

token_pause_transaction_e2e_test.go
```diff
@@ -0,0 +1,111 @@
...
+		SetTokenID(tokenID).
+		Execute(env.Client)
+
+	require.NoError(t, err)
+
+	err = CloseIntegrationTestEnv(env, &tokenID)
+	require.NoError(t, err)
+}
+
+func TestIntegrationTokenPauseTransactionWrongPauseKey(t *testing.T) {
```
bamzedev, line 66: Not sure we need this as an E2E test. Simple unit test with `toProto`, `fromProto` should be enough.

bamzedev (changes requested):

token_pause_transaction_e2e_test.go
```diff
@@ -0,0 +1,111 @@
+package hedera
```
bamzedev, line 1:
```
//go:build all || e2e
// +build all e2e
```

token_pause_transaction_e2e_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+import (
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
```
bamzedev, line 5: maybe we use just `assert` or just `require`

bamzedev:
The commit is not signed

bamzedev (approved):

PR #765, opened 2023-06-28
Fixed network increase backoff concurrency issue

### LINKED ISSUE #761: Out of bound access in _Network._GetNodeAccountIDsForExecute
When getting the healthy nodes to build a transaction, there is no bound check on line 134-135.

This would rarely cause issue when the `Client` is created with default to pick 1/3 of nodes and majority of the nodes are healthy.

However, if a user sets a much higher number through `Client.SetMaxNodesPerTransaction` or creates a `Client` with a much smaller set of nodes, the chances to hit the out of bound access is much higher.   

https://github.com/hashgraph/hedera-sdk-go/blob/f6a3c6f3720b45f6b077c156e5957db385dc539a/network.go#L131-L136

https://github.com/hashgraph/hedera-sdk-go/blob/f6a3c6f3720b45f6b077c156e5957db385dc539a/managed_network.go#L143-L150

### LINKED ISSUE #762: _Network._IncreaseBackoff may accidentally remove healthy node
### Description

Since there's no multi-thread protection from goroutines calling _Network.IncreaseBackoff concurrently, for a specific node, when the function is called, it might have been already removed from the `healthyNodes` slice. However, `index` is initialized to 0, then the first node (different than the node to remove) gets removed from the `healthyNodes` slice.

https://github.com/hashgraph/hedera-sdk-go/blob/f6a3c6f3720b45f6b077c156e5957db385dc539a/network.go#L66-L80

### Steps to reproduce

refer to the description

### Additional context

_No response_

### Hedera network

other

### Version

main

### Operating system

None

rokn:
**Description**:
This PR fixes some concurrency issues with the Network instances, related to accessing `healthyNodes`. It's protecting only the part of the instance which is used when executing transactions.

**Related issue(s)**:
Fixes #762
Fixes #761 

**Checklist**

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,7 +12,7 @@ on:
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
   cancel-in-progress: true
-    
+
 jobs:
     build:
         name: Build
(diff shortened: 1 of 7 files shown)


xin-hedera:
suggest: add `-race` flag to `go test` in the github workflow, that should help detect potential data race issues

bamzedev (approved):

Issue #764, opened 2023-06-26
Various issues when interacting with smart contracts with integers


bamzedev:
### Description

Often transactions just get reverted when sending Ints to smart contracts. It seems the encoding is just not correct. It also can't send or receive negative values.

### Steps to reproduce

Deploy a contract that accepts Int. Try to send negative value, same goes for receiving.

### Hedera network

mainnet

### Version

v2.24.0

### Operating system

None

Issue #701, opened 2023-03-06
Get test coverage to 60%


bamzedev:
### Problem

Test coverage is too low.

### Solution

Get the coverage to 60%

PR #767, opened 2023-07-05
Fix: Contract function parameters

bamzedev:
**Description**:
This PR fixes the issues that the SDK have with different types when interacting with smart contracts.
It has breaking changes as some functions are now returning different types, i.e. instead of uint now return int.

Maybe we need to create a new "wrapper" functions for the uint/int types as the current solution is very hard for the users to deal with.
**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-sdk-go/issues/764, https://github.com/hashgraph/hedera-sdk-go/issues/769, https://github.com/hashgraph/hedera-sdk-go/issues/701

**Notes for reviewer**:
Currently some tests fail due to parallel execution. Will work on that.

**Checklist**

- [ x ] Documented (Code comments, README, etc.)
- [ x ] Tested (unit, integration, etc.)


contract_function_parameters.go
```diff
@@ -1159,13 +1766,19 @@ func (contract *ContractFunctionParameters) AddBytesArray(value [][]byte) *Contr
 func (contract *ContractFunctionParameters) AddBytes32Array(value [][]byte) *ContractFunctionParameters {
 	argument := _NewArgument()
 	argument.dynamic = true
+	// Each item is 32 bytes. The total size should be len(value) * 32
```
rokn, line 1769: nit:
```suggestion
	// Each item is 32 bytes. The total size should be len(value) * 32 plus 32 bytes for the length header
```

contract_function_parameters.go
```diff
@@ -1159,13 +1766,19 @@ func (contract *ContractFunctionParameters) AddBytesArray(value [][]byte) *Contr
...
-	result := make([]byte, len(value)+32)
-
+	// Write the length of the array into the first 32 bytes
 	binary.BigEndian.PutUint64(result[24:32], uint64(len(value)))
 
 	for i, v := range value {
-		copy(result[i*32+32:i*32+32+32], v[0:32])
+		// Ensure that each byte slice is 32 bytes long.
+		var b [32]byte
+		copy(b[:], v)
```
rokn, line 1778: Why do we need this instead of `v[:32]`?

contract_function_parameters.go
```diff
@@ -1159,13 +1766,19 @@ func (contract *ContractFunctionParameters) AddBytesArray(value [][]byte) *Contr
...
+
+		// Then copy into the result
+		copy(result[i*32+32:(i+1)*32+32], b[:])
```
rokn, line 1781: nit: can we extract the indices into a `from` and `to` vars for readability?

contract_function_parameters.go
```diff
@@ -115,7 +146,7 @@ func (contract *ContractFunctionParameters) AddInt8(value int8) *ContractFunctio
 
 // AddInt16 adds an int16 parameter to the function call
 func (contract *ContractFunctionParameters) AddInt16(value int16) *ContractFunctionParameters {
-	argument := _NewArgument()
+	argument := _NewInt16Argument(value)
 
 	binary.BigEndian.PutUint16(argument.value[30:32], uint16(value))
```
rokn, line 151: I'm not sure, wouldn't these need to be two's complemented as well

contract_function_result.go
```diff
@@ -232,6 +236,35 @@ func (result ContractFunctionResult) GetInt256(index uint64) []byte {
 	return result.ContractCallResult[index*32 : index*32+32]
 }
 
+func toBigIntFromTwosComplement(data []byte) *big.Int {
```
rokn, line 239: Here again I think we can use https://pkg.go.dev/github.com/ethereum/go-ethereum/common/math#S256

contract_function_result.go
```diff
@@ -414,6 +447,19 @@ func (result ContractFunctionResult) AsBytes() []byte {
 	return result.ContractCallResult
 }
 
+func (result ContractFunctionResult) GetResult(types string) (interface{}, error) {
```
rokn, line 450: Can we get a doc here as it's not instantly clear what this does

contract_function_parameters.go
```diff
@@ -597,7 +928,19 @@ func (contract *ContractFunctionParameters) AddUint72(value []byte) *ContractFun
 func (contract *ContractFunctionParameters) AddUint80(value []byte) *ContractFunctionParameters {
 	argument := _NewArgument()
 
-	argument.value = append(argument.value[:22], value...)
+	argument.value = append(argument.value[:0], value...)
```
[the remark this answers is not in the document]
bamzedev: left over from testing phase

contract_function_parameters.go
bamzedev: `v[:32]` does not always result in a 32byte, in which case it will panic later.

bamzedev: Fixed to left padding.
`copy(b[32-len(v):], v)`

rokn (approved):

Issue #758, opened 2023-06-21
NewContractCreateFlow does not work with some contracts 


bamzedev:
### Description

```hedera.NewContractCreateFlow().SetBytecode(bytecode).SetGas(10000000).Execute(client)``` returns `INSUFFICIENT_GAS` on testnet and `INVALID_SOLIDITY_ADDRESS` on previewnet.
This is the contract that reproduces it:
```// SPDX-License-Identifier: GPL-3.0-or-later
pragma solidity 0.8.4;

contract Storage {
    uint256 number;

    function storeUint256(uint256 num) public {
        number = num;
    }

    function storeAndRetrieveUint256(uint256 num) public returns (uint256) {
        number = num;
        return number;
    }

    function retrieve() public view returns (uint256) {
        return number;
    }
}

### Steps to reproduce

`hedera.NewContractCreateFlow().SetBytecode(bytecode).SetGas(10000000).Execute(client)`

### Additional context

_No response_

### Hedera network

testnet, previewnet

### Version

v2.24.0

### Operating system

None

PR #779, opened 2023-07-28
Fix: ContractCreateFlow

bamzedev:
**Description**:
ContractCreateFlow was failing/passing depending on whether the user passed hex/ascii and also size.
It now works with ascii and does not depend on the size of the contract.

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-sdk-go/issues/758

**Checklist**


rokn (approved):

Coiling-Dragon (approved):
LGTM

Issue #821, opened 2023-10-27
Setting nodes for transactions is wrong


rokn:
### Description

When creating a new transaction we always set the same nodes for this transaction. This should be done in a randomized manner. 

### Steps to reproduce

1. Create multiple transactions
2. All of them have the same first healthy nodes

### Hedera network

mainnet, testnet, previewnet

### Version

v0.31.0

### Operating system

None

PR #828, opened 2023-11-07
Fix/wrong nodes for transactions

bamzedev:
**Description**:
This PR adds a randomization on nodes, so that it does not always picks the same node.

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-sdk-go/issues/821

**Checklist**

### DIFF
--- a/network.go
+++ b/network.go
@@ -21,6 +21,7 @@ package hedera
  */
 
 import (
+	"math/rand"
 	"time"
 )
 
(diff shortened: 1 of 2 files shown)


network.go
```diff
@@ -143,6 +144,12 @@ func (network *_Network) _GetNodeAccountIDsForExecute() []AccountID { //nolint
 		nodes = append(nodes, network.healthyNodes[i].(*_Node).accountID)
 	}
 
+	// shuffle the nodes, so that the first node is not always the same
+	for i := range nodes {
+		j := rand.Intn(i + 1) // #nosec
+		nodes[i], nodes[j] = nodes[j], nodes[i]
+	}
+
```
rokn, line 152: I think we should randomize the whole list before getting the 1/3rd of the nodes. So we can hit all of the nodes, not just the first third.

rokn (approved):

PR #862, opened 2024-01-31
861 hip 745 integration tests

NicolaMirchev:
**Description**:
Added more integration tests regarding HIP-745

Fixes #

**Notes for reviewer**:

**Checklist**

- [x ] Tested (unit, integration, etc.)


gsstoykov (changes requested):
I think that the integration tests for example for AccountCreateTransaction should belong to the integration tests of the specific transaction not having a separate `serialize_deserialize_test.go`. Also do some investigation on `FileAppendTransactions` there should be a logic for `ChunkedTransaction` and add the same tests there if not all of them just demonstrate the old flow and then add a Serialize->Deserialize->Modify the transaction fields a bit and then compare.

gsstoykov (approved):
LGTM.

PR #879, opened 2024-02-20
feat(Go-TCK): Implement `JSON-RPC` methods for `AccountCreateTransaction`, `AccountUpdateTransaction`, `AccountDeleteTransaction` and `AccountInfoQuery`

### LINKED ISSUE #874: `TCK-Go`: Implement JSON-RPC methods for `AccountCreateTransaction`
We should implement JSON-RPC methods (endpoints) for `AccountCreateTransaction` for the Go TCK server

**Add the following methods:**

Account:
- createAccount
- createAccountFromAlias

### Note:
- **createAccount**: 
    - `should set account auto renew period to 2,592,000 seconds`
    - `Max token set to the maximum`

- **createAccountFromAlias**:
    - `Create account with alias`

### LINKED ISSUE #875: `TCK-Go`: Implement JSON-RPC methods for `AccountUpdateTransaction`
We should implement JSON-RPC methods (endpoints) for `AccountUpdateTransaction` for the Go TCK server

**Add the following methods:**

Account:
- updateAccount

@notice 
`updateAccountMemo` and `updateAccountMemo` should be refactored and combined in one single method

### Note:
- `updateAccountKey`
    - `Creates an account and sets staked account ID with no input`
 
- `updateAccountMemo`
    - `should test memo field is too long`

### LINKED ISSUE #876: `TCK-Go`: Implement JSON-RPC methods for `AccountDeleteTransaction`
We should implement JSON-RPC methods (endpoints) for `AccountDeleteTransaction` for the Go TCK server

**Add the following methods:**

Account:
- deleteAccount

### LINKED ISSUE #877: `TCK-Go`: Implement JSON-RPC methods for `AccountInfoQuery`
We should implement JSON-RPC methods (endpoints) for `AccountInfoQuery` for the Go TCK server

__Add the following methods:__
- getAccountInfo

agadzhalov:
**Description**:
Implement `JSON-RPC` methods for `AccountCreateTransaction`, `AccountUpdateTransaction`, `AccountDeleteTransaction` and `AccountInfoQuery`

**Related issue(s)**:

Fixes #874 
Fixes #875 
Fixes #876 
Fixes #877


tck/cmd/server.go
```diff
@@ -15,12 +18,23 @@ func main() {
 
 	// Initialize the SDK service
 	sdkService := new(methods.SDKService)
+	accountService := new(methods.AccountService)
```
gsstoykov, line 21: We can have a lazier initialisation for these services. If this function is responsible for any service initialisation it would scale up in code and even guessing runtimes.

tck/methods/account.go
```diff
@@ -0,0 +1,279 @@
...
+	Key       string `json:"key"`
+	Memo      string `json:"memo"`
+}
+type DeleteAccountParams struct {
+	AccountId   string `json:"accountId"`
+	AccountKey  string `json:"accountKey"`
+	RecipientId string `json:"recipientId"`
+}
+
+// TODO: Move this struct in dedicated file for hepler structs, or something like this. We want to use it for all methods
```
gsstoykov, line 54: Different file with Error definitions would be appropriate here.

tck/methods/account.go
```diff
@@ -0,0 +1,279 @@
...
+	AccountID                     string
+	Balance                       string
+	Key                           string
+	AccountMemo                   string
+	MaxAutomaticTokenAssociations uint32
+	AutoRenewPeriod               time.Duration
+}
+
+// Variable to be set to `SetGrpcDeadline` for all transactions
+var threeSec = time.Second * 3
```
gsstoykov, line 74: Would remove this variable.

tck/methods/account.go
```diff
@@ -0,0 +1,279 @@
...
+	if accountParams.ReceiverSignatureRequired {
+		transaction.SetReceiverSignatureRequired(accountParams.ReceiverSignatureRequired)
+	}
+	if accountParams.MaxAutomaticTokenAssociations != 0 {
+		transaction.SetMaxAutomaticTokenAssociations(accountParams.MaxAutomaticTokenAssociations)
+	}
+	if accountParams.StakedAccountId != "" {
+		accountId, err := hedera.AccountIDFromString(accountParams.StakedAccountId)
+		if err != nil {
+			return nil, jrpc2.Errorf(-32603, "Internal error", ErrorStatus{Status: err.Error()})
```
gsstoykov, line 106: If the JSON RPCs have a standard error code like -32603 would define it somewhere. Will improve readability.

tck/methods/account.go
```diff
@@ -0,0 +1,279 @@
...
+	resp, err := tx.Sign(signature).Execute(a.sdkService.Client)
+	if err != nil {
+		return nil, jrpc2.Errorf(-32603, "Internal error", ErrorStatus{Status: err.Error()})
+	}
+
+	receipt, _ := resp.GetReceipt(a.sdkService.Client)
+	return &AccountResponse{Status: receipt.Status.String()}, nil
+}
+
+// Helper methods
```
gsstoykov, line 234: These methods are public and have the same naming convention as the above methods which are intended for use from the TCK server. Would make them private or at least make them follow a different naming convention

tck/methods/account.go
```diff
@@ -0,0 +1,279 @@
...
+	receipt, err := resp.GetReceipt(a.sdkService.Client)
+	if err != nil {
+		return nil, jrpc2.Errorf(-32603, "Internal error", ErrorStatus{Status: err.Error()})
+	}
+
+	return &AccountResponse{Status: receipt.Status.String()}, nil
+}
+
+// UpdateAccountMemo updates account memo of an existing account ID
+func (a *AccountService) UpdateAccountMemo(accountId string, key string, memo string) (*AccountResponse, error) {
```
gsstoykov, line 264: I see no reason why these should take multiple args. Would just pass the parameters of the caller function.

tck/methods/key.go
```diff
@@ -0,0 +1,33 @@
+package methods
+
+import (
+	"context"
+	"github.com/hashgraph/hedera-sdk-go/v2"
+)
+
+type KeyService struct {
+}
```
gsstoykov, line 9:
```
type AccountService struct {
	sdkService *SDKService
	hedera.AccountID
}
```
Shouldn't this be defined in similar fashion?

tck/methods/key.go
```diff
@@ -0,0 +1,33 @@
...
+func (_ KeyService) GeneratePublicKey(_ context.Context, params KeyParams) string {
+	key, err := hedera.PrivateKeyFromString(params.PrivateKey)
+	if err != nil {
+		return ""
+	}
+	return key.PublicKey().String()
+}
+
+// GeneratePrivateKey generate new private key and return it.
+// NOTE: params object is used only to maintain the communication with jRPC client, which is providing empty object when calling this func
```
gsstoykov, line 25: Incorrect comment. The params have nothing to do with jRPC clients. I guess that `context` was assumed here. That would make me think that there is no specific need to pass `context` everywhere. Can we wrap the functions so `context` isn't passed everywhere?

tck/methods/account.go
```diff
@@ -0,0 +1,279 @@
...
+	return nil, nil
+}
+
+// DeleteAccount deletes a provided account by signing the transaction with the key of that account
+func (a *AccountService) DeleteAccount(_ context.Context, params DeleteAccountParams) (*AccountResponse, error) {
+	accId, _ := hedera.AccountIDFromString(params.AccountId)
+	recipientId, _ := hedera.AccountIDFromString(params.RecipientId)
+	tx, err := hedera.NewAccountDeleteTransaction().SetGrpcDeadline(&threeSec).SetAccountID(accId).SetTransferAccountID(recipientId).FreezeWith(a.sdkService.Client)
+	if err != nil {
+		return nil, jrpc2.Errorf(-32603, "Internal error", ErrorStatus{Status: err.Error()})
```
gsstoykov, line 221: Would guess that `Internal error` will not be the only received error. Should standardise a bit more around error definitions in the TCK projects.

tck/methods/account.go
```diff
@@ -0,0 +1,279 @@
...
+		AutoRenewPeriod:               resp.AutoRenewPeriod}, nil
+}
+
+type UpdateAccountParams struct {
+	AccountId string `json:"accountId"`
+	// Params for updateAccountKey
+	NewPublicKey  string `json:"newPublicKey"`
+	OldPrivateKey string `json:"oldPrivateKey"`
+	NewPrivateKey string `json:"newPrivateKey"`
+	// Params for updateAccountMemo
```
gsstoykov, line 199: No need to specify this. We could have endpoints that would need another variation. Comments regarding the purpose of the parameters here are irrelevant.

tck/methods/account.go
```diff
@@ -0,0 +1,279 @@
...
+// SetSdkService We set object, which is holding our client params. Pass it by referance, because TCK is dynamically updating it
+func (a *AccountService) SetSdkService(service *SDKService) {
+	a.sdkService = service
+}
+
+// TODO Find better way to handle error. We want some wrapper, which would intercept the error and contruct the correct
+// TODO jrpc error based on that  "jrpc2.Errorf(-32603, "Internal error"/"Hedera error", ErrorStatus{Status: err.Error()})"
+// CreateAccount jRPC method for createAccount
+func (a *AccountService) CreateAccount(_ context.Context, accountParams CreateAccountParams) (*AccountResponse, error) {
+	transaction := hedera.NewAccountCreateTransaction().SetGrpcDeadline(&threeSec)
```
gsstoykov, line 85: No need to have reference to three seconds.

tck/methods/account.go
```diff
@@ -0,0 +1,279 @@
...
+	var accId string
+	if receipt.Status == hedera.StatusSuccess {
+		accId = receipt.AccountID.String()
+	}
+
+	return &AccountResponse{AccountId: accId, Status: receipt.Status.String()}, nil
+}
+
+// GetAccountInfo Get info for a given accountId and return a custom object containing aggregated information about the account
+func (a *AccountService) GetAccountInfo(_ context.Context, accountId string) (*AccountInfoResponse, error) {
```
gsstoykov, line 177: Would pass parameters here no need to have different standards for function calls.

tck/methods/account.go
```diff
@@ -0,0 +1,279 @@
...
+	}
+
+	resp, err := hedera.NewTransferTransaction().SetGrpcDeadline(&threeSec).
+		AddHbarTransfer(operator, hedera.NewHbar(float64(fromAliasParams.InitialBalance)).Negated()).
+		AddHbarTransfer(alias, hedera.NewHbar(float64(fromAliasParams.InitialBalance))).
+		Execute(a.sdkService.Client)
+
+	if err != nil {
+		fmt.Println(err)
+		return nil, jrpc2.Errorf(-32603, "Internal error").WithData(ErrorStatus{Status: err.Error()})
```
gsstoykov, line 165: In some places `WithData` is used in some places it is done like:
```
if err != nil {
		return nil, jrpc2.Errorf(-32603, "Internal error", ErrorStatus{Status: err.Error()})
```

tck/methods/account.go
```diff
@@ -0,0 +1,279 @@
...
+		_, err = transaction.FreezeWith(a.sdkService.Client)
+		if err != nil {
+			return nil, jrpc2.Errorf(-32603, err.Error())
+		}
+		transaction.Sign(key)
+	}
+
+	txResponse, err := transaction.Execute(a.sdkService.Client)
+	if err != nil {
+		fmt.Println(err)
```
gsstoykov, line 137: Guessing this came for some sort of debugging. Should be removed.

gsstoykov: There are others in this file.

tck/methods/account.go
gsstoykov: We can find multiple occurances of this.

tck/methods/account.go
```diff
@@ -0,0 +1,279 @@
...
+	AliasAccountId string `json:"aliasAccountId"`
+	InitialBalance int64  `json:"initialBalance"`
+}
+type UpdateAccountKeyParams struct {
+	AccountId     string `json:"accountId"`
+	NewPublicKey  string `json:"newPublicKey"`
+	OldPrivateKey string `json:"oldPrivateKey"`
+	NewPrivateKey string `json:"newPrivateKey"`
+}
+type UpdateAccountMemoParams struct {
```
gsstoykov, line 43: Not used anywhere. And not needed as well. We can directly pass params.

gsstoykov (changes requested):

tck/cmd/server.go
```diff
@@ -29,8 +43,16 @@ func main() {
...
-		panic(err)
-	}
+	fmt.Println("Server is listening on port: " + port)
+	go func() {
+		http.ListenAndServe(":"+port, nil)
+	}()
+
+	signalCh := make(chan os.Signal, 1)
+	signal.Notify(signalCh, os.Interrupt, syscall.SIGTERM)
+	sig := <-signalCh
```
0xivanov, line 53: Could add timeout context of 5 sec to the server so that any ongoing requests are finished before shutting down the server.

```go
    ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
	defer cancel()

	if err := server.Shutdown(ctx); err != nil {
		log.Printf("Graceful shutdown failed: %v\n", err)
		os.Exit(1)
	}
```

tck/methods/account.go
```diff
@@ -0,0 +1,279 @@
...
+// GetAccountInfo Get info for a given accountId and return a custom object containing aggregated information about the account
+func (a *AccountService) GetAccountInfo(_ context.Context, accountId string) (*AccountInfoResponse, error) {
+	account, _ := hedera.AccountIDFromString(accountId)
+
+	resp, err := hedera.NewAccountInfoQuery().SetAccountID(account).SetGrpcDeadline(&threeSec).Execute(a.sdkService.Client)
+	if err != nil {
+		return nil, jrpc2.Errorf(-32603, "Internal error", ErrorStatus{Status: err.Error()})
+	}
+
+	return &AccountInfoResponse{AccountID: resp.AccountID.String(),
```
gsstoykov, line 185: Only `accountID` is required to verify the test.

tck/methods/account.go
0xivanov: Refactored the error handling

tck/methods/account.go
0xivanov: I tried passing the value but it does not work, i think the variable is needed

tck/methods/account.go
0xivanov: Done

tck/methods/account.go
0xivanov: Done

tck/methods/account.go
0xivanov: Done

tck/methods/key.go
0xivanov: The keys generation does not make requests to the network, so it does not need client.

tck/methods/account.go
0xivanov: Made it so we return the error and the errorMiddleware handles the specific error logic.

tck/methods/account.go
```diff
@@ -0,0 +1,234 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	receipt, _ := resp.GetReceipt(a.sdkService.Client)
+	return &response.AccountResponse{Status: receipt.Status.String()}, nil
+}
+
+// UpdateAccountKey updates an existing acoount id keys with provided params
+func (a *AccountService) UpdateAccountKey(params param.UpdateAccountParams) (*response.AccountResponse, error) {
```
0xivanov, line 173: make private

tck/methods/account.go
```diff
@@ -0,0 +1,234 @@
...
+	receipt, err := resp.GetReceipt(a.sdkService.Client)
+	if err != nil {
+		return nil, err
+	}
+
+	return &response.AccountResponse{Status: receipt.Status.String()}, nil
+}
+
+// UpdateAccountMemo updates account memo of an existing account ID
+func (a *AccountService) UpdateAccountMemo(params param.UpdateAccountParams) (*response.AccountResponse, error) {
```
0xivanov, line 213: make private

tck/cmd/server.go
```diff
@@ -29,8 +48,66 @@ func main() {
...
+
+// ErrorHandler is a function that handles errors reported by a method handler.
+type ErrorHandler func(context.Context, *jrpc2.Request, error) error
+
+func HandleError(_ context.Context, request *jrpc2.Request, err error) error {
 	if err != nil {
-		panic(err)
+		// jrpc generic error
+		if jrpcError, ok := err.(*jrpc2.Error); ok {
+			return jrpcError
```
0xivanov, line 88: return `NOT_IMPLEMENTED` if method not found line in js:
```js
    if (jsonRPCResponse.error) {
        if (jsonRPCResponse.error.code === -32601){
            console.warn("Method", method, "not found.")
            return {status: "NOT_IMPLEMENTED"}; // The json-rpc hasn't implemented the method
        }
        throw {name: "Error", ...jsonRPCResponse.error}
```

0xivanov: Nevermind, this is handled in the tck driver itself.

tck/methods/account.go
gsstoykov: Not sure it should be left private since the TCK has a specified test for `UpdateAccountKey`. Not fully sure tho.

tck/methods/account.go
gsstoykov: Not sure it should be left private since the TCK has a specified test for `UpdateAccountMemo`. Not fully sure tho.

gsstoykov (approved):
LGTM.

tck/methods/account.go
0xivanov: Yes, you are right. We should add these methods to the assigner map.

go.work.sum
```diff
@@ -0,0 +1,18 @@
+github.com/deckarep/golang-set v1.8.0 h1:sk9/l/KqpunDwP7pSjUg0keiOOLEnOBHzykLrsPppp4=
```
0xivanov, line 1: Should probably remove this file

tck/methods/account.go
0xivanov, line 173: Add this method to the assigner map and test if it works with these params.

go.work.sum
gsstoykov: Yeah it seems old. It's holding the old version of the protobufs.. If not you can try to update it but just remove IMHO.

0xivanov: Done

tck/methods/account.go
0xivanov: Done

tck/cmd/server.go
```diff
@@ -15,12 +24,23 @@ func main() {
...
 	// Initialize the SDK service
 	sdkService := new(methods.SDKService)
+	accountService := new(methods.AccountService)
+	accountService.SetSdkService(sdkService)
 
 	// Create a new RPC server
 	assigner := handler.Map{
-		"setup": handler.New(sdkService.Setup),
-		"reset": handler.New(sdkService.Reset),
+		"setup":                  errorMiddleware(HandleError, handler.New(sdkService.Setup)),
```
gsstoykov, line 32: Looks like we are initialising the server with errorMiddleware's and not endpoints and etc. Would think of renaming this but anyways this is a NIT.

0xivanov: Renamed

PR #920, opened 2024-05-09
feat: HIP-540 Change Or Remove Existing Keys From A Token

### LINKED ISSUE #776: Add HIP-540: Change Or Remove Existing Keys From A Token
### Problem

HIP-540 introduces new Hedera protobuf changes to TokenUpdateTransaction where the user will be able to remove keys that were initially set on the token. 

**HIP**

- [x] https://hips.hedera.com/hip/hip-540

**Hedera Protobuf**

- [ ] (link Hedera protobuf PR when available)

### Solution

- [ ] Design the feature
     - [ ] Add to the [SDK references repository](https://github.com/hashgraph/hedera-sdk-reference)
- [ ] Implement design
- [ ] Add example

### Alternatives

_No response_

0xivanov:
**Description**:
This PR implements HIP-540.

**Related issue(s)**:

Fixes #776

**Notes for reviewer**:

**Checklist**


.github/workflows/build.yml
```diff
@@ -58,7 +58,7 @@ jobs:
 
             - name: Start the local node
               if: success()
-              run: npx @hashgraph/hedera-local start -d --network local
+              run: npx @hashgraph/hedera-local start -d --network local --network-tag=0.50.0-alpha.0
```
gsstoykov, line 61: Should this be against a specific network tag or just against hedera-local@latest?

token_update_transaction_e2e_test.go
```diff
@@ -843,3 +843,212 @@ func TestIntegrationTokenUpdateTransactionCannotUpdateImmutableNFT(t *testing.T)
...
+	receipt, err = resp.SetValidateStatus(true).GetReceipt(env.Client)
+	require.NoError(t, err)
+
+}
+
+func TestIntegrationTokenUpdateTransactionUpdateAdminKeyWithoutAlreadySetAdminKeyFails(t *testing.T) {
+	t.Parallel()
+	env := NewIntegrationTestEnv(t)
+
+	zeroAdminKey, err := PrivateKeyFromString("0000000000000000000000000000000000000000000000000000000000000000")
```
gsstoykov, line 883: nit: Maybe you can somehow define the zeroAdminKey string. It looks bad having all that zeros in the tests.

token_update_transaction_e2e_test.go
```diff
@@ -843,3 +843,212 @@ func TestIntegrationTokenUpdateTransactionCannotUpdateImmutableNFT(t *testing.T)
...
+	require.NoError(t, err)
+
+	tx, err := NewTokenUpdateTransaction().
+		SetTokenID(*receipt.TokenID).
+		SetAdminKey(zeroAdminKey).
+		SetKeyVerificationMode(NO_VALIDATION).
+		FreezeWith(env.Client)
+	assert.NoError(t, err)
+
+	// Sign with the supply key should fail
```
gsstoykov, line 941: nit: If you have a comment here you can go ahead and add comments for the above steps as well. This would apply for other functions as well.

gsstoykov (commented):
Overall LGTM. Few nits and maybe we can add more tests from the test plan and would be ready for merge.

token_update_transaction_e2e_test.go
0xivanov: Added comments

token_update_transaction_e2e_test.go
0xivanov: added zeroKeyString const

gsstoykov (approved):
All looks good but should be merged after Mirror Node Queries and tests rerun.

token_update_transaction_hip540_e2e_test.go
```diff
@@ -0,0 +1,718 @@
...
+}
+
+func updateTokenKeysHelper(t *testing.T, tokenID TokenID, updateKeyType KeyType, client *Client, newKey Key, signerKey PrivateKey, verificationMode TokenKeyValidation) (TransactionResponse, error) {
+	privateKey, _ := newKey.(PrivateKey)
+	// Update the key
+	tx2 := NewTokenUpdateTransaction().
+		SetTokenID(tokenID).
+		SetKeyVerificationMode(verificationMode)
+
+	switch updateKeyType {
```
gsstoykov, line 641: Maybe the switch can be reused in a function taking in a transaction and a the KeyType

gsstoykov: Also the NONE value is never checked against in the switch

jeromy-cannon (approved):

token_update_transaction_hip540_e2e_test.go
```diff
@@ -0,0 +1,718 @@
...
+	require.NoError(t, err)
+
+	resp, err = updateTokenKeysHelper(t, *receipt.TokenID, updateKeyType, client, newKey, signerKey, verificationMode)
+	return resp, *receipt.TokenID, err
+}
+
+func updateTokenKeysHelper(t *testing.T, tokenID TokenID, updateKeyType KeyType, client *Client, newKey Key, signerKey PrivateKey, verificationMode TokenKeyValidation) (TransactionResponse, error) {
+	privateKey, _ := newKey.(PrivateKey)
+	// Update the key
+	tx2 := NewTokenUpdateTransaction().
```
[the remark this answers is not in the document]
0xivanov: fixed

jeromy-cannon (dismissed):

token_update_transaction_hip540_e2e_test.go
```diff
@@ -0,0 +1,718 @@
...
+	SUPPLY_KEY
+	FREEZE_KEY
+	FEE_SCHEDULE_KEY
+	PAUSE_KEY
+	METADATA_KEY
+	ADMIN_KEY
+	ALL
+	LOWER_PRIVILEGE
+	NONE
+	ALL_DIFFERENT_ADMIN_KEY
```
gsstoykov, line 52: This is never used.

token_update_transaction_hip540_e2e_test.go
```diff
@@ -0,0 +1,718 @@
...
+)
+
+const zeroKeyString = "0000000000000000000000000000000000000000000000000000000000000000"
```
gsstoykov, line 55: Maybe we should note to users that they can use this zeroKeyString from the HIP specification. The only place you can see this in action is this integration test suite. Maybe we can move the key to some other file in C++ for example I did something like this: 
```
/**
   * SDK needs to provide  a way to set an unusable key such as an Ed25519 all-zeros
   * key, since it is (presumably) impossible to find the 32-byte string whose SHA-512 hash begins with 32 bytes
   * of zeros. We recommend using all-zeros to clearly advertise any unsuable keys.
   */
  static inline const std::string ZERO_KEY_STR = "0000000000000000000000000000000000000000000000000000000000000000";
``` In the ED25519PrivateKey class.

File: token_update_transaction_hip540_e2e_test.go (hunk 6 above)
Comment (0xivanov) [reply token_update_transaction_hip540_e2e_test.go]: Removed

File: token_update_transaction_hip540_e2e_test.go (hunk 7 above)
Comment (0xivanov) [reply token_update_transaction_hip540_e2e_test.go]: Added const ZERO_KEY_STRING

**Review by 0xivanov — COMMENTED** [review]
Should revert the default local node version before merging.

File: crypto.go [hunk 8]
```diff
@@ -91,6 +91,15 @@ type PublicKey struct {
 	ed25519PublicKey *_Ed25519PublicKey
 }
 
+/**
+ *  SDK needs to provide  a way to set an unusable key such as an Ed25519 all-zeros
+ *  key, since it is (presumably) impossible to find the 32-byte string whose SHA-512 hash begins with 32 bytes
+ *  of zeros. We recommend using all-zeros to clearly advertise any unusable keys.
+ */
+func UnusableKey() (PublicKey, error) {
```
gsstoykov, line 99: nit: Would just avoid using unusable as the key name it might bring confusion. Instead lets leave it zeroKey or something similar.

gsstoykov (dismissed):

crypto.go
0xivanov: Renamed

gsstoykov (approved):

Issue #841, opened 2023-12-20
Add Account Balance/Info, Contract Info Mirror Node Queries 


NicolaMirchev:
### Problem

Detailed: https://github.com/hashgraph/hedera-sdk-reference/issues/144

### Solution

Add a service to query balance data for the corresponding network:

Detailed:
https://www.notion.so/limechain/Mirror-Node-Queries-fe0d563db0d545fdafd7a0a5137fe5bd

HIP: https://hips.hedera.com/hip/hip-873

Mirror Node: https://github.com/hashgraph/hedera-mirror-node/issues/7386

agadzhalov:
Work is continued on separated issues

Go-TCK[: Implement JSON-RPC methods for ](https://github.com/hashgraph/hedera-sdk-go/issues/874)AccountCreateTransaction
Go-TCK: [Implement JSON-RPC methods for](https://github.com/hashgraph/hedera-sdk-go/issues/875) AccountUpdateTransaction
Go-TCK[: Implement JSON-RPC methods for ](https://github.com/hashgraph/hedera-sdk-go/issues/876)AccountDeleteTransaction
Go-TCK[: Implement JSON-RPC methods for ](https://github.com/hashgraph/hedera-sdk-go/issues/877)AccountInfoQuery

PR #930, opened 2024-05-14
feat(HIP-367) Mirror Node Queries

gsstoykov:
This PR introduces the SDK capability for querying `AccountInfo`, `ContractInfo`, `AccountBalance` and `TokenRelationships` from the Mirror Node. Currently not all Mirror Node queries will be used by the SDK but the `AccountInfo` and `TokenRelationship` queries are needed in order to populate the SDK fields for `TokenBalances` and `TokenRelationships` thus resolving the deprecations described in https://github.com/hashgraph/hedera-sdk-reference/issues/144.

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-sdk-go/issues/841

**Notes for reviewer**:
Currently the GO SDK has no straightforward way of verifying JSON objects. A note for future implementations would be adding a more sophisticated JSON validation.

At some point we should discuss approaches for testing the Mirror Node responses. Currently the responses are only validated through the flows of the integration tests. All tests including `AccountInfo`, `ContractInfo`, `AccountBalance` and `TokenRelationships` serve as validation of the queries.

Tests and examples are now dependant on Mirror Node update times so data from the queries will be assumed to be valid after a 2-3 second delay. Currently there is no way for working around that delay time as it is dependant on the Mirror Node itself.

**Checklist**


mirror_node_gateway.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+// Function to obtain account info for given account ID. Return the pure JSON response as mapping
+func AccountInfoMirrorNodeQuery(networkUrl string, accountId string) (map[string]interface{}, error) {
+	accountInfoUrl := buildUrlParams(networkUrl, "accounts", accountId)
+	return makeGetRequest(accountInfoUrl)
+}
+
+// Function to obtain balance of tokens for given contract ID. Return the pure JSON response as mapping
+func ContractInfoMirrorNodeQuery(networkUrl string, contractId string) (map[string]interface{}, error) { // nolint
```
0xivanov, line 43: This is not used

mirror_node_gateway.go
```diff
@@ -0,0 +1,99 @@
...
+	if strings.HasPrefix(client.GetMirrorNetwork()[0], localNetwork) {
+		return "http://" + localNetwork + ":5551" + apiVersion
+	} else {
+		// prefix is mainnet, testnet or previewnet
+		return "https://" + client.GetMirrorNetwork()[0] + apiVersion
+	}
+}
+
+// Make a GET HTTP request to provided URL and map it's JSON response to a generic `interface` map and return it
+func makeGetRequest(networkUrl string) (response map[string]interface{}, e error) {
```
0xivanov, line 71: I don't really like that we are returning generic `interface`. One suggestion - to modify your function to return the response body contents directly and let the caller handle the JSON decoding.

```go
// Make a GET HTTP request to provided URL and return its body and any potential error
func makeGetRequest(networkUrl string) (body []byte, err error) {
    // Make an HTTP request
    resp, err := http.Get(networkUrl)
    if err != nil {
        return nil, err
    }
    defer resp.Body.Close()

    // Read the response body
    body, err = ioutil.ReadAll(resp.Body)
    if err != nil {
        return nil, err
    }

    // Check for non-200 status codes
    if resp.StatusCode >= 400 {
        return nil, fmt.Errorf("HTTP request failed with status code: %d", resp.StatusCode)
    }

    return body, nil
}
```

0xivanov: We could ditch the `TokenRelationshipFromJson` and other similar methods if we are using `json.Unmarshal`

mirror_node_gateway.go
gsstoykov: Yeah, we had it defined at some point when we saw the specification differently. But I don't think we should remove it because if the mirror node query gateway where to be a full implementation it should match the api defined here:
https://mainnet-public.mirrornode.hedera.com/api/v1/docs/#/

mirror_node_gateway.go
```diff
@@ -0,0 +1,99 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"strings"
+)
+
+// Function to obtain the token relationships of the specified account
+func TokenReleationshipMirrorNodeQuery(networkUrl string, id string) (map[string]interface{}, error) {
```
0xivanov, line 31: Typo

```suggestion
func TokenRelationshipMirrorNodeQuery(networkUrl string, id string) (map[string]interface{}, error) {
```

mirror_node_gateway.go
0xivanov: Could be done when we introduce "full" queries to the mirror node

mirror_node_gateway.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+// Function to obtain balance of tokens for given contract ID. Return the pure JSON response as mapping
+func ContractInfoMirrorNodeQuery(networkUrl string, contractId string) (map[string]interface{}, error) { // nolint
+	contractInfoUrl := buildUrlParams(networkUrl, "contracts", contractId)
+	return makeGetRequest(contractInfoUrl)
+}
+
+// Function to obtain balance of tokens for given account ID. Return the pure JSON response as mapping
+func AccountBalanceMirrorNodeQuery(networkUrl string, accountId string) (map[string]interface{}, error) {
```
0xivanov, line 49: Can be renamed to `AccountTokenBalance` since we want to get only the token balances.

0xivanov: Also here we are using `AccountInfoQuery` but in java sdk it's using the `TokenRelationship` query to get the token balances. 
https://github.com/hashgraph/hedera-sdk-java/pull/1698/files#diff-edc9172a6752893870b903e7c87f7227511b3820fcf6e5d31d092c80dd940afcR117

mirror_node_gateway.go
0xivanov: Currently we don't have test that is asserting the `TokenRelationships` from the token info. Should add.

account_balance.go
```diff
@@ -26,13 +26,10 @@ import (
...
 	Hbars Hbar
-
 	// Deprecated: Use `AccountBalance.Tokens` instead
-	Token map[TokenID]uint64
-
-	// Deprecated
+	Token  map[TokenID]uint64
 	Tokens TokenBalanceMap
-	// Deprecated
+	// Deprecated: use a mirror node query
```
0xivanov, line 32: Might be good idea to remove `Deprecated` for fields that are queried from mirror node. They are not actually being deprecated from the sdks

token_relationship.go
```diff
@@ -19,89 +19,60 @@ package hedera
...
  *
  */
+import (
+	"github.com/pkg/errors"
+)
 
 // TokenRelationship is the information about a token relationship
 type TokenRelationship struct {
 	TokenID              TokenID
-	Symbol               string
```
0xivanov, line 26: Revert

mirror_node_gateway.go
```diff
@@ -0,0 +1,99 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"strings"
+)
+
+// Function to obtain the token relationships of the specified account
+func TokenRelationshipMirrorNodeQuery(networkUrl string, id string) (map[string]interface{}, error) {
```
0xivanov, line 31: Make these private

0xivanov (approved):
LGTM

PR #947, opened 2024-06-03
fix: Raise an error if the transaction is not frozen while signing

### LINKED ISSUE #946: `Sign` without `Freeze` does not fail
### Description

Signing requires the transaction to be frozen. We need to make sure the request is frozen before signing. Add check for this and propagate `errTransactionIsNotFrozen` if the check fails

### Steps to reproduce

Sign and execute transaction without freezing.

### Additional context

_No response_

### Hedera network

testnet

### Version

2.37.0

### Operating system

None

0xivanov:
**Description**:

Signing requires the transaction to be frozen. We need to make sure the request is frozen before signing. Add check for this and propagate `errTransactionIsNotFrozen` if the check fails.

**Related issue(s)**:

Fixes #946

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/account_create_transaction_e2e_test.go
+++ b/account_create_transaction_e2e_test.go
@@ -355,12 +355,14 @@ func TestIntegrationAccountCreateTransactionWithAliasFromAdminKeyWithReceiverSig
 		Execute(env.Client)
 	require.NoError(t, err)
 
-	resp, err := NewAccountCreateTransaction().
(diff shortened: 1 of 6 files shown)


gsstoykov (approved):
LGTM. Just a note:
Looks like HIP-745 features are poorly tested. Maybe we should have a follow-up on this.

PR #950, opened 2024-06-04
fix: Account Alias Mirror Node Queries

### LINKED ISSUE #948: Fix mirror node queries when using account alias
### Description

The mirror node does not accept der encoded account aliases. We cannot query newly created hollow account info and balanced because of this.

We should use the protobuf response in order to get the num of the account/contract and use it as parameter for the query.

### Steps to reproduce

1. Create hollow account
2. Set it's `AccountID` as query parameter.

### Additional context

_No response_

### Hedera network

previewnet

### Version

2.37.0

### Operating system

None

0xivanov:
**Description**:

The mirror node does not accept der encoded account aliases. We cannot query newly created hollow account info and balanced because of this.

We should use the protobuf response in order to get the num of the account/contract and use it as parameter for the query.

**Related issue(s)**:

Fixes #948

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/token_transfer_transaction_e2e_test.go
+++ b/token_transfer_transaction_e2e_test.go
@@ -26,6 +26,7 @@ package hedera
 import (
 	"fmt"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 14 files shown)


token_transfer_transaction_e2e_test.go
```diff
@@ -110,6 +111,9 @@ func TestIntegrationTokenTransferTransactionCanExecute(t *testing.T) {
 	_, err = resp.SetValidateStatus(true).GetReceipt(env.Client)
 	require.NoError(t, err)
 
+	// Wait for mirror node to import data
+	time.Sleep(3 * time.Second)
```
0xivanov, line 115: Remove this

gsstoykov (dismissed):
Lets do a couple of reruns for the integration tests so we are sure that flakiness is gone. Otherwise LGTM.

gsstoykov (approved):

gsstoykov (approved):

PR #986, opened 2024-06-20
fix: Properly handle `PlatformNotActive` status code

0xivanov:
**Description**:

Currently when `PlatformNotActive` code gets returned from consensus in transactions or queries, we retry. 
But this is not the case for `TransactionReceiptQuery` and `TransactionRecordQuery` because they have custom implementation for the `shouldRetry` method.

This PR fixes returning precheck error in case of `PlatformNotActive` status. If `PlatformNotActive` occurs, we should retry.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


gsstoykov (dismissed):

gsstoykov (approved):
LGTM.

Issue #893, opened 2024-03-19
Write unit tests for the go `TCK`


0xivanov:
### Problem

Write unit tests for the go `TCK`

### Solution

-

0xivanov:
fixed in #1012

PR #1012, opened 2024-07-30
fix(TCK): `AccountCreateTransaction`

### LINKED ISSUE #1013: fix(TCK): JSON-RPC methods for `AccountCreateTransaction`
### Problem

Fix AccountCreate tck method to work with the latest version of the tck server

### Solution

-

### Alternatives

_No response_

0xivanov:
**Description**:

Fix `AccountCreate` tck method to work with the latest version of the tck server

**Related issue(s)**:

Fixes #1013 #893

**Notes for reviewer**:

**Checklist**


transaction.go
```diff
@@ -4877,6 +4877,27 @@ func (tx *Transaction) execute(client *Client, e TransactionInterface) (Transact
 	}, nil
 }
 
+func (tx *Transaction) FreezeWith(client *Client, e TransactionInterface) (TransactionInterface, error) { //nolint
```
0xivanov, line 4880: TODO: we need a way of calling freeze on a transaction, without knowing the concrete type of the transaction. It's used in `FillOutTransaction` method

0xivanov (commented):
Adding some notes for reviewers and for myself.

crypto.go
```diff
@@ -32,6 +32,7 @@ import (
 	"github.com/hashgraph/hedera-protobufs-go/services"
 	"github.com/pkg/errors"
 	"golang.org/x/crypto/pbkdf2"
+	protobuf "google.golang.org/protobuf/proto"
 )
 
 const _Ed25519PrivateKeyPrefix = "302e020100300506032b657004220420"
```
gsstoykov, line 38: Note for the future: This is not actually correct.

key_list.go
```diff
@@ -73,7 +78,7 @@ func (kl *KeyList) AddAllPublicKeys(keys []PublicKey) *KeyList {
 }
 
 // String returns a string representation of the KeyList
```
gsstoykov, line 80: nit: Guess this is some old typo.

tck/cmd/server.go
```diff
@@ -29,16 +29,10 @@ func main() {
...
-		"reset":                  errorHandlerWrapper(HandleError, handler.New(sdkService.Reset)),
-		"createAccount":          errorHandlerWrapper(HandleError, handler.New(accountService.CreateAccount)),
-		"createAccountFromAlias": errorHandlerWrapper(HandleError, handler.New(accountService.CreateAccountFromAlias)),
-		"getAccountInfo":         errorHandlerWrapper(HandleError, handler.New(accountService.GetAccountInfo)),
-		"updateAccountMemo":      errorHandlerWrapper(HandleError, handler.New(accountService.UpdateAccountMemo)),
-		"updateAccountKey":       errorHandlerWrapper(HandleError, handler.New(accountService.UpdateAccountKey)),
-		"deleteAccount":          errorHandlerWrapper(HandleError, handler.New(accountService.DeleteAccount)),
-		"generatePublicKey":      errorHandlerWrapper(HandleError, handler.New(methods.GeneratePublicKey)),
-		"generatePrivateKey":     errorHandlerWrapper(HandleError, handler.New(methods.GeneratePrivateKey)),
+		"setup":         errorHandlerWrapper(HandleError, handler.New(sdkService.Setup)),
```
gsstoykov, line 32: We are binding the endpoints to `errorHandlerWrapper`. This doesn't look that good in my opinion. Would expect something like `setup: handler.New()`). Maybe we can rename to `HandleWrapper`.

transaction.go
gsstoykov: Could it be achievable if `FillOutTransaction` is a method of the Transactions themselves?

crypto.go
0xivanov: We have the same in JS

key_list.go
0xivanov: I don't see typo in the comment

transaction.go
0xivanov: I don't see the point, as `FillOutTransaction` would not fit in any way in the existing flows of the sdk. 
I just need a way to expose the freezeWith function, as it's exposed in Java.

crypto.go
gsstoykov: I guess so but for example `30740201010420` would still be a valid prefix if you integrate the key data, curve data and etc. after it. Thats why I put in this notice.

key_list.go
gsstoykov: Not a typo in the traditional sense but is String returns a string in the comment correct here?

transaction.go
gsstoykov: We can take a look. Maybe I don't have enough context.

key_list.go
0xivanov: I think `String` is referring to the name of the method

0xivanov (commented):
Had to force push the whole tck server, because the `feat/tck-poc` was behind 80 commits and it was a mess.

tck/cmd/server.go
0xivanov: Renamed this to `postHandler`. Have a look.

tck/cmd/server.go
```diff
@@ -0,0 +1,109 @@
...
+
+	bridge := jhttp.NewBridge(assigner, nil)
+
+	// Listen and redirect to bridge
+	http.HandleFunc("/", bridge.ServeHTTP)
+	port := os.Getenv("TCK_PORT")
+	if port == "" {
+		port = "80"
+	}
+	fmt.Println("Server is listening on port: " + port)
```
gsstoykov, line 46: Maybe at some point we can refactor to using log or logrus.

tck/methods/key_test.go
```diff
@@ -0,0 +1,182 @@
...
+		Threshold: 1,
+		Keys:      nil,
+	}
+
+	// When
+	_, err := GenerateKey(context.Background(), params)
+
+	// Then
+	assert.Error(t, err)
+	assert.Contains(t, err.Error(), "invalid parameters: threshold should only be provided for thresholdKey types")
```
gsstoykov, line 42: Maybe we should think of a way to centralise the error messages. As a changes for the error in one place for example the `GenerateKey` may lead to changes in the tests and overall test inconsistency. I personally don't like checking against concrete strings.

tck/methods/sdk_test.go
```diff
@@ -0,0 +1,72 @@
...
+
+	// When
+	response := sdkService.Reset(context.Background())
+
+	// Then
+	assert.Equal(t, "SUCCESS", response.Status)
+	assert.Nil(t, sdkService.Client)
+}
+
+func stringPointer(s string) *string {
```
gsstoykov, line 70: Would you need this if for example you had:
type SetupParams struct {
	OperatorAccountId  string  `json:"operatorAccountId"`
	OperatorPrivateKey string  `json:"operatorPrivateKey"`
	NodeIp             string `json:"nodeIp"`
	NodeAccountId      string `json:"nodeAccountId"`
	MirrorNetworkIp    string `json:"mirrorNetworkIp"`
}

tck/param/sdk.go
```diff
@@ -0,0 +1,9 @@
+package param
+
+type SetupParams struct {
```
gsstoykov, line 3: Is there are reason for some of these being pointers and others not?

0xivanov: Basically the ones that are being pointers are optional -> they can be nil.

0xivanov: I will try to make every param struct be like this.

tck/methods/key_test.go
0xivanov: Moved the errors to utils/key.go

tck/cmd/server.go
0xivanov: changed

tck/param/sdk.go
0xivanov: Done

tck/methods/sdk_test.go
0xivanov: changed to use pointers where optional

gsstoykov (approved):
LGTM.

agadzhalov (approved):
LGTM

PR #1038, opened 2024-08-26
feat: Simplified ipv4 address

gsstoykov:
**Description**:
Ipv4 implementation had a useless abstraction. 

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-sdk-go/issues/1036

**Notes for reviewer**:

**Checklist**


0xivanov (commented):
Looks good, needs some more manual testing if it does not break anything during actual address book update.

ipv4_address.go
```diff
@@ -21,47 +21,24 @@ package hedera
  */
 
 type IPv4Address struct {
```
0xivanov, line 23: I think this stuct can be removed as well, it just holds the `address []byte` and does not really do anything.
For example the `endpoint` should call `endpoint.address.address` which is weird.

gsstoykov: True, lets remove.

query.go
```diff
@@ -203,7 +203,7 @@ func (q *Query) SetPaymentTransactionID(transactionID TransactionID) *Query {
 
 func (q *Query) execute(client *Client, e QueryInterface) (*services.Response, error) {
 	q.client = client
-	if client == nil || client.operator == nil {
```
0xivanov, line 206: note: removed since some queries don't require operator.

account_balance_query_e2e_test.go
```diff
@@ -212,3 +214,31 @@ func TestIntegrationAccountBalanceQueryWorksWithHollowAccountAlias(t *testing.T)
...
+	client.SetTransportSecurity(true)
+
+	succeededOnce := false
+	for address, accountID := range client.GetNetwork() {
+
+		if !strings.HasSuffix(address, ":50212") {
+			t.Errorf("Expected entry key to end with ':50212', but got %s", address)
+		}
+
+		fmt.Println(address, accountID)
```
0xivanov, line 230: can be removed

0xivanov (approved):
LGTM

PR #1070, opened 2024-09-17
feat(HIP-904): Token Claim/Cancel

### LINKED ISSUE #1032: Add HIP-904: `TokenClaimAidropTranasction`

0xivanov:
**Description**:

**Related issue(s)**:

Fixes #1032 #1031 #1027

**Notes for reviewer**:

**Checklist**


utilities_for_test.go
```diff
@@ -323,3 +323,5 @@ func createFungibleToken(env *IntegrationTestEnv, opts ...TokenCreateTransaction
 	}
 	return *receipt.TokenID, err
 }
+
+const transferAmount = 100
```
gsstoykov, line 327: I think that values specific to a test should be defined in the tests themselves. I like the idea we have on reusing certain repeated transactions but I think this could potentially grow into having lots of specific constants in the utils for testing.

token_claim_airdrop_transaction_e2e_test.go
```diff
@@ -0,0 +1,544 @@
...
+	_, err = claimResp.SetValidateStatus(true).GetReceipt(env.Client)
+	require.ErrorContains(t, err, "INVALID_PENDING_AIRDROP_ID")
+}
+
+func TestIntegrationTokenClaimAirdropCannotClaimWithEmptyPendingAirdrops(t *testing.T) {
+	env := NewIntegrationTestEnv(t)
+	defer CloseIntegrationTestEnv(env, nil)
+
+	// Claim the tokens with the receiver without setting pendingAirdropIds
+	// Fails with EMPTY_PENDING_AIRDROP_ID_LIST
```
gsstoykov, line 376: nit: I think this is obvious but you can leave it as is just marking. Would only lead to a risk of mismatch of the comment and the actual string type if modified by services.

gsstoykov (commented):
LG overall. Just a little comment for the test utilities.

utilities_for_test.go
0xivanov: Moved

token_claim_airdrop_transaction_e2e_test.go
0xivanov: Removed

PR #1071, opened 2024-09-18
release: v2.46.0

0xivanov:
**Description**:

 - release: v2.46.0
 - update address book


addressbook/mainnet.pb
```diff
@@ -620,4 +620,285 @@
```
gsstoykov, line 620: This concerns me a bit having it now. And this is a release PR. Maybe move the addressbook update for some later task or if someone requires it. Wouldn't risk braking any user here even if the chance is really slight.

gsstoykov (commented):
Is the addressbook update intentional?

PR #1089, opened 2024-10-02
refactor: Implement generics in `Transaction.go`

0xivanov:
**Description**:

## Introduction

The current design of the `Transaction` struct in Go has 3 major inefficiencies:

1. **Redundant method shadowing**: Each transaction type (e.g., `AccountCreateTransaction`, `TokenCreateTransaction`, etc.) needs to implement boilerplate functions that simply shadow methods from the base `Transaction` struct. This results in thousands of lines of redundant code and poor maintainability as the number of methods increases. 
2. **Complex static utility functions**: Functions such as `TransactionExecute` and `TransactionSign` rely on large switch statements to handle various transaction types. As more transaction types are added, the code becomes increasingly complex, leading to maintainability issues and a higher likelihood of errors. For reference, the current lines of code for these static methods is around 4000. 
One such function is this:
3. Adding new abstractions to handle similar transaction types (`Airdrop/Transfer`, `TokenClaim/TokenCancel`) is costly in terms of code duplication and complexity. Every transaction type needs its own implementation, which makes code reuse difficult.

In order to fix all above mentioned problems is by implementing Generics into `Transacation.go` and transactions. 

# Overview

### Pros:

1. **Implements the Fluent Pattern**:
    - The current SDK design uses the fluent interface, making it easy for developers to chain method calls and maintain a clean, readable API.
     Example:  `NewTokenCreateTransaction.SetTokenID().SetNodeAccountIDs().Execute()`

### Cons:

1. **Bad Scaling**:
    - The SDK suffers from method shadowing, which leads to thousands of lines of redundant code (currently around 5000 lines). As the number of methods grows, managing these shadowed methods becomes more troublesome, leading to poor scalability.
2. **Costly to Create Intermediate Abstractions**:
    - Adding new abstractions to handle similar transaction types is costly in terms of code duplication and complexity. Every transaction type needs its own implementation, which makes code reuse difficult.
3. **Hard to Maintain**:
    - The base transaction class is coupled with many concrete subclasses (~100). Introducing new methods in the base class requires updating all the subclasses, creating a maintenance burden.
4. **Prone to Errors - No Compile-time Checks**:
    - The current design lacks type safety at compile time, meaning many potential errors can only be caught at runtime. This increases the risk of bugs that are harder to track down.

---

### Pros:

1. **Implements the Fluent Pattern**:
    - Like the old architecture, the new architecture retains the fluent interface, ensuring that the API remains developer-friendly for building complex chains of setters.
    Example : `NewTokenCreateTransaction.SetTokenID().SetNodeAccountIDs().Execute()`
2. **Good Scaling**:
    - By introducing a more reusable base transaction, many common transaction types can share the same base logic. This makes the codebase much more scalable and avoids the redundancy issues of the old design.
3. **Easy to Maintain**:
    - The base transaction can be extended with additional common fields and methods, making it much easier to modify or enhance functionality without needing to touch every subclass. This significantly reduces the maintenance burden.
4. **Compile-time Checks**:
    - The new design leverages Go's type system to enforce stronger compile-time checks, reducing the risk of runtime errors and improving overall code reliability.
5. **Easier to Create Intermediate Abstractions**:
    - The new architecture allows more flexibility when creating abstractions. Shared logic can be factored out into reusable components, reducing the amount of duplicate code.

### Cons:

1. Needs one change to the existing api:
    - `SignTransaction` ’s api needs a minor change. It would need to accept concrete transaction instead of parent transaction.
    Example before: `newKey.SignTransaction(&transaction.Transaction)`
    Example after: `newKey.SignTransaction(transaction)`
        - I’ve done some research and this method is not commonly used, but if someone is already consuming this API, it can be changed very easily (only remove the `.Transaction` at the end).

---

https://www.notion.so/limechain/Go-SDK-refactor-v2-10610baa8dcd80d7a787d086a3ef3181

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


crypto.go
```diff
@@ -860,9 +860,11 @@ func (pk PublicKey) _ToSignaturePairProtobuf(signature []byte) *services.Signatu
 }
 
 // SignTransaction signes the transaction and adds the signature to the transaction
-func (sk PrivateKey) SignTransaction(tx *Transaction) ([]byte, error) {
```
0xivanov, line 863: Changed API

file_append_transaction.go
```diff
@@ -468,12 +346,16 @@ func (tx *FileAppendTransaction) buildProtoBody() *services.FileAppendTransactio
...
 
-func (tx *FileAppendTransaction) getMethod(channel *_Channel) _Method {
+func (tx FileAppendTransaction) getMethod(channel *_Channel) _Method {
 	return _Method{
 		transaction: channel._GetFile().AppendContent,
 	}
 }
 
-func (tx *FileAppendTransaction) _ConstructScheduleProtobuf() (*services.SchedulableTransactionBody, error) {
+func (tx FileAppendTransaction) constructScheduleProtobuf() (*services.SchedulableTransactionBody, error) {
```
0xivanov, line 355: Would like to remove this func as it's useless.

transaction.go
```diff
@@ -782,49 +823,183 @@ func (tx *Transaction) _BuildTransaction(index int) (*services.Transaction, erro
 // Shared
 //
 
-// GetMaxTransactionFee returns the maximum transaction fee the operator (paying account) is willing to pay.
-func (tx *Transaction) GetMaxTransactionFee() Hbar {
-	return HbarFromTinybar(int64(tx.transactionFee))
+// GetSignedTransactionBodyBytes
+// TODO remove this method, it's not used in the sdk and it does not seem useful for the user
+func (tx *Transaction[T]) GetSignedTransactionBodyBytes(transactionIndex int) []byte {
```
0xivanov, line 828: Would like to remove since it's unused and it's not present functionality in other sdks.

transaction.go
```diff
@@ -782,49 +823,183 @@ func (tx *Transaction) _BuildTransaction(index int) (*services.Transaction, erro
...
+func (tx *Transaction[T]) GetTransactionMemo() string {
 	return tx.memo
 }
 
 // SetTransactionMemo sets the memo for this transaction.
-func (tx *Transaction) SetTransactionMemo(memo string) *Transaction {
+func (tx *Transaction[T]) SetTransactionMemo(memo string) T {
 	tx.memo = memo
-	return tx
+	// tx.executable.transactionIDs = _NewLockableSlice()
```
0xivanov, line 997: Remove

0xivanov (commented):
Do not merge until we've release v2.49

account_update_transaction_e2e_test.go
```diff
@@ -229,9 +229,9 @@ func TestIntegrationAccountUpdateTransactionAccountIDNotSet(t *testing.T) {
 //	updateBytes, err := tx.ToBytes()
 //	require.NoError(t, err)
 //
-//	sig1, err := newKey.SignTransaction(&tx.Transaction)
+//	sig1, err := newKey.SignTransaction(tx.Transaction)
```
gsstoykov: nit: Just changes on commented code.

transaction.go
```diff
@@ -96,6 +98,153 @@ func (tx *Transaction[T]) GetSignedTransactionBodyBytes(transactionIndex int) []
 	return tx.signedTransactions._Get(transactionIndex).(*services.SignedTransaction).GetBodyBytes()
 }
 
+// Helper function to cast any type of transaction to the base transaction
+func castFromAnyToBaseTransaction(tx any) (*Transaction[TransactionInterface], error) { // nolint
```
gsstoykov: If all transaction type cases call `castFromConcreteToBaseTransaction(t)` why do we have this switch case?

gsstoykov (approved):
LGTM.

transaction_generics_utils.go
gsstoykov, line 1: nit: I don't think this file should exist. Just hold the logic in transaction.go.

0xivanov: Removed

gsstoykov (approved):

Issue #1113, opened 2024-10-23
replace `/crypto` from `go-ethereum`


0xivanov:
### Problem

replace `/crypto` from `go-ethereum`

### Solution

-

PR #1124, opened 2024-10-31
chore: Replace key logic from `go-ethereum`

0xivanov:
**Description**:
Remove `/crypto` package coming from go-ethereum.

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-sdk-go/issues/1113

**Notes for reviewer**:

**Checklist**


query.go
```diff
@@ -170,6 +170,9 @@ func _QueryMakePaymentTransaction(transactionID TransactionID, nodeAccountID Acc
 	}
 
 	signature := operator.signer(bodyBytes)
+	if len(signature) == 65 {
```
0xivanov, line 173: note:
The ecdsa signature is 65 bytes normally. For some reason the consensus node requires it to be 64 bytes. go-ethereum also produces 65 bytes sig and we used to cut it like this in the `sign` func: 
```go
return sig[:len(sig)-1]
```

Even though we used to cut it, the `VerifySignature` func used to work, but now it does not - if we cut the signature in the `sign` method. That's why the caller of the `sign` method should cut it with the introduction of the new lib.

gsstoykov: 64-byte ECDSA signature: Contains only the r and s values, sufficient for standard cryptographic verification. 65-byte ECDSA signature: Contains r, s, and a 1-byte v (recovery id) value, commonly used to allow recovery of the public key directly from the signature. This extra byte is specific to certain implementations and is not part of the core ECDSA signature format itself, but rather an application-specific addition. 
TLDR; maybe consensus doesn't need the `recId`

gsstoykov (approved):
LGTM.

0xivanov (commented):
Do not merge, until /abi and /rlp PRs are merged.

gsstoykov (approved):

Issue #1111, opened 2024-10-23
Replace `/common/abi` from `go-ethereum`


0xivanov:
### Problem

Replace `/common/abi` from `go-ethereum`

### Solution

-

PR #1127, opened 2024-11-05
chore: replace `/common/abi` module

0xivanov:
**Description**:
Replace `/common/abi` from go-ethereum

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-sdk-go/issues/1111

**Notes for reviewer**:

**Checklist**


contract_function_parameters_e2e_test.go
```diff
@@ -319,1247 +317,1249 @@ func performDeploy(env IntegrationTestEnv) (*ContractID, error) {
...
 	require.NoError(t, err)
 	require.Equal(t, value, contractCal.GetUint8(0))
-
+	err = CloseIntegrationTestEnv(env, nil)
+	require.NoError(t, err)
 }
 func TestUint8Max(t *testing.T) {
 
 	env := NewIntegrationTestEnv(t)
-	defer CloseIntegrationTestEnv(env, nil)
```
0xivanov, line 334: revert the defer changes

abi.go
```diff
@@ -0,0 +1,446 @@
...
+
+	for _, field := range fields {
+		switch field.Type {
+		case "constructor":
+			if a.Constructor != nil {
+				return fmt.Errorf("multiple constructor declaration")
+			}
+			input, err := NewTupleTypeFromArgs(field.Inputs)
+			if err != nil {
+				panic(err)
```
gsstoykov, line 147: Should we panic here? Why not just propagate an error?

abi.go
```diff
@@ -0,0 +1,446 @@
...
+			}
+			a.Constructor = &Method{
+				Inputs: input,
+			}
+
+		case "function", "":
+			c := field.Constant
```
gsstoykov, line 154: Do we need `field.Constant` if it is just used to assign `c` and `c` is always the result of `field.StateMutability == "view" || field.StateMutability == "pure"`

abi_decode.go
```diff
@@ -0,0 +1,427 @@
...
+	return a
+}
+
+// Address implements the ethgo.Key interface Address method.
+func (a Address) Address() Address {
+	return a
+}
+
+// Sign implements the ethgo.Key interface Sign method.
+func (a Address) Sign(hash []byte) ([]byte, error) {
```
gsstoykov, line 162: If `an address cannot sign messages` why have this function?

abi_encode.go
```diff
@@ -0,0 +1,377 @@
...
+	one  = big.NewInt(1)
+)
+
+// Encode encodes a value
+func Encode(v interface{}, t *Type) ([]byte, error) {
+	return encode(reflect.ValueOf(v), t)
+}
+
+func encode(v reflect.Value, t *Type) ([]byte, error) {
+	if v.Kind() == reflect.Interface {
```
gsstoykov, line 43: nit: Valid here and for every `kind`. Won't it be better named as something like `dataType`, `valueType`?

ethereum_transaction_e2e_test.go
gsstoykov, line 1: note: There is going to be a merge conflict here with the new RLP impl. It will no longer require this decodeHex. Just pointing out.

abi_encode.go
```diff
@@ -0,0 +1,377 @@
...
+// nolint
+func leftPad(b []byte, size int) []byte {
+	return padBytes(b, size, true)
+}
+
+func rightPad(b []byte, size int) []byte {
+	return padBytes(b, size, false)
+}
+
+func decodeHex(str string) ([]byte, error) {
```
gsstoykov, line 370: I think this name is too general to be package private in the hedera package. I suggest renaming it to a more specific name.

ethereum_transaction_e2e_test.go
gsstoykov: If decodeHex in you other file gets renamed to a more specific name there won't be a conflict here,

gsstoykov (commented):
Overall LG, just some final grooming suggestions.

ethereum_transaction_e2e_test.go
0xivanov: Not sure what this means since i cannot see the outlined code.

abi_encode.go
0xivanov: Im open for suggestions.

abi_encode.go
0xivanov: Line 43 is checking `kind` which is from the go standard lib. The `kinds` in the switch statement are solidity types. I've renamed the solidity `Kind` to `AbiTypeKind` to make it clearer.

abi.go
0xivanov: fixed.

abi_decode.go
0xivanov: removed.

abi_encode.go
gsstoykov: Maybe you can leave it as is and delete the decodeHex in the `ethereum_transaction` test(Assuming it will do the same). This way we won't have to change the test and no merge conflict will follow.

0xivanov: The test function panics, while the actual implementation is returning an error. I made the test function an internal variable, this should fix the compilation error.

0xivanov: The test util func is removed.

ethereum_transaction_e2e_test.go
0xivanov: Fixed the merge conflicts

gsstoykov (approved):

Issue #1114, opened 2024-10-23
replace `/rlp` from `go-ethereum`


0xivanov:
### Problem

replace `/rlp` from `go-ethereum`

### Solution

-

PR #1128, opened 2024-11-06
feat: Replace go-ethereum RLP lib

gsstoykov:
**Description**:
Introduces new RLP decoding/encoding logic to substitute the `go-ethereum` provided lib. Also introduces new custom Ethereum transaction types that where previously used from `go-ethereum`.

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-sdk-go/issues/1114

**Notes for reviewer**:

**Checklist**


ethereum_eip1559_transaction.go
```diff
@@ -0,0 +1,167 @@
...
+		hex.EncodeToString(txn.CallData),
+		accessListStr,
+		hex.EncodeToString(txn.RecoveryId),
+		hex.EncodeToString(txn.R),
+		hex.EncodeToString(txn.S),
+	)
+}
+
+// Helper function to create an RLP item for value assignment
+func (item *RLPItem) EIP1559TransactionAssignBytes(value []byte) *RLPItem {
```
0xivanov, line 163: This is the same as `LegacyTransactionAssignBytes` right?
If so - I think it can be moved to `rlp.go` (since this func belongs to `RLPItem`) and rename it to something like `func (item *RLPItem) AssignBytes(value []byte) *RLPItem {}`

ethereum_transaction_data.go
```diff
@@ -20,121 +20,61 @@ package hedera
...
 		return ethereumTxData
 	}
 
-	ethereumTxData.legacy.Data = data
+	ethereumTxData.legacy.CallData = data
 	return ethereumTxData
 }
-
-// ToJson returns a JSON representation of the Ethereum transaction.
-func (ethereumTxData *EthereumTransactionData) ToJson() ([]byte, error) {
```
0xivanov, line 101: We should check if these json related funcs were not requested from user and if it's safe to remove them.

ethereum_eip1559_transaction.go
0xivanov: Now i see that we already have `AssignBytes` func. Maybe you can make it return an `*RLPItem` and it would do the job, i think.

0xivanov (commented):
Looks pretty good. Licenses are missing in some files.

ethereum_transaction_data.go
```diff
@@ -20,121 +20,61 @@ package hedera
...
 		if err != nil {
 			return nil, err
 		}
 
+		transactionData.eip1559 = eip1559
 		return &transactionData, nil
 	}
 
-	err := rlp.DecodeBytes(b, &transactionData.legacy)
+	legacy, err := EthereumLegacyTransactionFromBytes(b)
```
0xivanov, line 42: The legacy transaction flows are not tested. Could add similar tests for the legacy, as eip1559

ethereum_transaction_data.go
0xivanov: Do the new struct types break the to/fromJson funcs? The old implementation looks like it would work.

ethereum_transaction_e2e_test.go
```diff
@@ -128,36 +82,23 @@ func TestIntegrationEthereumTransaction(t *testing.T) {
...
-
-	bytes, _ := rlp.EncodeToBytes(list)
 
-	// 02 is the type of the transaction EIP1559 and should be concatenated to the RLP by service requirement
-	bytesToSign := append(decodeHex(t, "02"), bytes...)
-	signedBytes := ecdsaPrivateKey.Sign(bytesToSign)
+	// signedBytes := ecdsaPrivateKey.Sign(bytesToSign)
+
+	// dummy signed data until test is revisitted
+	rlp, err := hex.DecodeString("02f87082012a022f2f83018000947e3a9eaf9bcc39e2ffa38eb30bf7a93feacbc181880de0b6b3a764000083123456c001a0df48f2efd10421811de2bfb125ab75b2d3c44139c4642837fb1fccce911fd479a01aaf7ae92bee896651dfc9d99ae422a296bf5d9f1ca49b2d96d82b79eb112d66")
```
0xivanov, line 89: Can we create the `RLPList` with all it's properties, like it was done before this change. Maybe you can create a simple contract in remix IDE and see some of the properties from there.

I just don't like this string - it's hard to see what's going on.

ethereum_transaction_data.go
0xivanov: Will create a follow-up ticket for this.

ethereum_transaction_e2e_test.go
0xivanov: Will create a follow up ticket for this.

0xivanov (approved):
LGTM

PR #1135, opened 2024-11-13
fix: Retry in case of throttle when calling `getRecord`

### LINKED ISSUE #1125: Fix `THROTTLED_AT_CONSENSUS` for `getRecord`
### Description

`getRecord` can sometimes cause `THROTTLED_AT_CONSENSUS` status and we need to handle it.

### Steps to reproduce

TBD

### Additional context

_No response_

### Hedera network

_No response_

### Version

latest

### Operating system

None

0xivanov:
**Description**:

**Related issue(s)**:

Fixes #1125

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/transaction_response.go
+++ b/transaction_response.go
@@ -91,6 +91,10 @@ func (response TransactionResponse) GetRecord(client *Client) (TransactionRecord
 		SetNodeAccountIDs([]AccountID{response.NodeID}).
 		Execute(client)
 
+	for receipt.Status == StatusThrottledAtConsensus {
+		receipt, err = retryTransaction(client, response.Transaction)
+	}
+
 	if err != nil {
 		// Manually add the receipt, because an empty TransactionRecord will have an empty receipt and empty receipt has no status and no status defaults to 0, which means success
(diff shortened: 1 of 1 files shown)


gsstoykov (approved):
Looks good. We can rerun the CI integration tests a couple of times just to be safe.

transaction_response.go
```diff
@@ -91,6 +91,10 @@ func (response TransactionResponse) GetRecord(client *Client) (TransactionRecord
 		SetNodeAccountIDs([]AccountID{response.NodeID}).
 		Execute(client)
 
+	for receipt.Status == StatusThrottledAtConsensus {
+		receipt, err = retryTransaction(client, response.Transaction)
+	}
+
```
gsstoykov, line 97: Maybe this could be verified with:
Creating a test which constantly tries to get a record of a transaction. But could be a loss of CI time as we have to bake a certain amount of time which we would constantly loop over a transaction being sent and getting its record? WDYT?

0xivanov: Yeah, this would slow down a lot CI time, since the throttles are quite random.

PR #1137, opened 2024-11-13
feat: Init client with mirror network

### LINKED ISSUE #1126: Add option to automatically query for an `AddressBook` if a `Client` is initialized only with a `mirrorNetwork`
To allow Sphere users to fully initialize a `Client` with only a mirror node address, there should be an option that allows for the initialization of the `Client`'s consensus node network automatically with an `AddressBookQuery` to the mirror node.

0xivanov:
**Description**:

**Related issue(s)**:

Fixes #1126

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/client.go
+++ b/client.go
@@ -81,6 +81,20 @@ var mainnetMirror = []string{"mainnet-public.mirrornode.hedera.com:443"}
 var testnetMirror = []string{"testnet.mirrornode.hedera.com:443"}
 var previewnetMirror = []string{"previewnet.mirrornode.hedera.com:443"}
 
+// ClientForMirrorNetwork constructs a client given a set of mirror network nodes.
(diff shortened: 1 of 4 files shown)


client.go
```diff
@@ -81,6 +81,20 @@ var mainnetMirror = []string{"mainnet-public.mirrornode.hedera.com:443"}
 var testnetMirror = []string{"testnet.mirrornode.hedera.com:443"}
 var previewnetMirror = []string{"previewnet.mirrornode.hedera.com:443"}
 
+// ClientForMirrorNetwork constructs a client given a set of mirror network nodes.
+func ClientForMirrorNetwork(mirrorNetwork []string) (*Client, error) {
```
0xivanov, line 85: note:
Does not work with local node.

gsstoykov (approved):
LGTM.

note:
This could be tested with Solo when the time comes

PR #1141, opened 2024-11-18
feat: create client without schedule network update

### LINKED ISSUE #1130: Unacceptable performance issue affecting versions 2.19.0 up to the latest making the SDK unusable for a mobile wallet
### Description

Per this article https://hedera.com/blog/deprecation-of-the-insecure-hedera-consensus-service-hcs-mirror-node-endpoints I am trying to get Wallawallet (IOS version) updated ASAP in the IOS app store before next week's deadline, yet when I tried the latest Hedera go sdk build (v2.49.0) it takes 44 seconds to fetch the balance and token balances from my IOS app and 4 minutes to generate a recovery phrase using the SDKs methods. The version of the SDK I am upgrading from (v2.17.6) from took less than 3 seconds, and there is no issue with 2.18.0. So the problem was introduced in version 2.19.0.

This is not a problem with query responses, but rather seems to be an issue with blocking the main thread from the limited amount of time I have had to look at this. I'm guessing perhaps there are some background init tasks that are blocking the main thread? Maybe checking node statuses, etc.? As a result, the user is left hanging for very long periods of time.

I am using React Native (contact me privately to learn versions being used if needed), go and gomobile. I changed nothing but my go library that is built with a Hedera sdk dependency. I also feed my own json config to it.

Please advise how I am to meet the deadline next week for mirror node/SDK changes as I am unable to find a working version of the SDK that is in the acceptable range of SDKs defined in the Hedera blog article referenced above. I would appreciate a quick fix and an extension of the deadline so that my app can be properly tested before going into production.

If needed I can provide some Wallawallet Test Flight builds to share/test versions of Wallawallet with any particular build you may recommend that meets the upcoming requirements.

Thanks

### Steps to reproduce

* Create a React Native mobile app, using the React Native bridge with gomobile and go code to call v2.49.0 Hedera go SDK methods from React Native.
* Launch the app on an iPhone
* Invoke the method for generating the random seed words or try to get your balance

### Additional context

_No response_

### Hedera network

mainnet

### Version

v2.49.0

### Operating system

Other

0xivanov:
**Description**:

Adding a new api that if a JSON config is provided there is no need for the address book update thread.

**Related issue(s)**:

Fixes #1130

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/client.go
+++ b/client.go
@@ -84,7 +84,7 @@ var previewnetMirror = []string{"previewnet.mirrornode.hedera.com:443"}
 // ClientForMirrorNetwork constructs a client given a set of mirror network nodes.
 func ClientForMirrorNetwork(mirrorNetwork []string) (*Client, error) {
 	net := _NewNetwork()
-	client := _NewClient(net, mirrorNetwork, nil)
+	client := _NewClient(net, mirrorNetwork, nil, true)
 	addressbook, err := NewAddressBookQuery().
 		SetFileID(FileIDForAddressBook()).
 		Execute(client)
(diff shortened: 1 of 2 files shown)


gsstoykov (approved):
LGTM.

Issue #1133, opened 2024-11-12
feat(TCK): Implement `TokenCreate` method


0xivanov:
### Problem

-

### Solution

-

PR #1162, opened 2024-11-25
feat(TCK): Implement token create method

0xivanov:
Implement token create method for the tck server.

Fixes: #1133


tck/methods/token.go
```diff
@@ -0,0 +1,264 @@
...
+				feeCollector, err := hiero.AccountIDFromString(customFee.FeeCollectorAccountId)
+				if err != nil {
+					return nil, err
+				}
+				fee.SetFeeCollectorAccountID(feeCollector)
+				fee.SetAllCollectorsAreExempt(*customFee.FeeCollectorsExempt)
+				customFeeList = append(customFeeList, fee)
+			}
+
+			// Handle Royalty Fee
```
gsstoykov, line 216: If we expect to have `customFee`s in other transactions like `TokenFeeScheduleUpdateTransaction` or potentially other places maybe we can reuse the marshalling logic.

0xivanov: nice catch, done

gsstoykov (approved):
LGTM.

PR #1174, opened 2024-12-05
chore: Update protobufs to `v0.57`

0xivanov:
This PR
- Updates the protobufs to v0.57
- Updates the protoc-go-gen version to latest, because the previous version is not compatible with go version 21


gsstoykov (approved):

gsstoykov (commented):
Services version could be bumped to `0.57.0` no need for the alpha

gsstoykov (approved):

PR #1188, opened 2024-12-13
feat: Mirror Node Contract Queries

### LINKED ISSUE #1167: Add a query to estimate the gas of a contract call
### Description
Hedera's EVM equivalence enables developers to deploy and execute smart contracts using standard EVM-compatible tools. Additionally, the Hedera SDK allows the execution of smart contract transactions (state-changing and query-based) via protobuf messages submitted to Consensus Nodes.

However, unlike typical EVM flows where transactions are encoded with RLP and submitted to a JSON-RPC endpoint, Hedera’s SDK can introduce friction for developers transitioning from other ecosystems.

### Problem

Today, the SDK does not support a way to determine the gas of a contract call. 

EVM developers, accustomed to estimating gas requirements as part of their workflow, face challenges using the SDK. Specifically:

- The `ContractCall` transaction type allows setting a gas limit but does not support gas estimation.
- This absence of gas estimation is a core gap, making the SDK less developer-friendly for those familiar with Ethereum-like environments.

### Solution

Add a query that estimates the gas for a contract call. 

1. Add a new query type
2. Add `getGasCost` Support to `ContractCall`

### Alternatives

_No response_

0xivanov:
**Description**:

A new `MirrorNodeContractQuery` class has been introduced to support EVM gas estimation and simulation. This query class provides two distinct types of queries:

1. **`MirrorNodeContractCallQuery`** – Executes a contract call and returns the result, simulating the execution of the contract.
2. **`MirrorNodeContractEstimateGasQuery`** – Estimates the gas usage for a contract call, allowing developers to understand the cost of executing a transaction.

The inputs for these queries match those of `ContractExecuteTransaction`, making it easy for developers to calculate the cost of a follow-up transaction.

The `MirrorNodeContractQuery` issues an HTTP request to the `api/v1/contracts/call` endpoint of the Mirror Node Web3 module, which supports both gas estimation and transient simulation of state-changing transactions. This enables accurate cost predictions and helps developers optimize their contract interactions before executing them on the network.

**Related issue(s)**:

Fixes #1167

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/account_id.go
+++ b/account_id.go
@@ -324,9 +324,6 @@ const (
 	EvmAddress
 )
 
-var httpString string = "http"
-var httpsString string = "https"
-
 func (id *AccountID) _MirrorNodeRequest(client *Client, populateType string) (map[string]interface{}, error) {
 	if client.mirrorNetwork == nil || len(client.GetMirrorNetwork()) == 0 {
 		return nil, errors.New("mirror node is not set")
@@ -340,11 +337,11 @@ func (id *AccountID) _MirrorNodeRequest(client *Client, populateType string) (ma
 	mirrorUrl = mirrorUrl[:index]
 
 	var url string
-	protocol := httpsString
+	protocol := "https"
 	port := ""
 
 	if client.GetLedgerID().String() == "" {
-		protocol = httpString
+		protocol = "http"
 		port = ":5551"
 	}
 
--- a/account_id.go
+++ b/account_id.go
@@ -324,6 +324,9 @@ const (
 	EvmAddress
 )
 
+var httpString string = "http"
+var httpsString string = "https"
+
 func (id *AccountID) _MirrorNodeRequest(client *Client, populateType string) (map[string]interface{}, error) {
 	if client.mirrorNetwork == nil || len(client.GetMirrorNetwork()) == 0 {
 		return nil, errors.New("mirror node is not set")
@@ -337,11 +340,11 @@ func (id *AccountID) _MirrorNodeRequest(client *Client, populateType string) (ma
 	mirrorUrl = mirrorUrl[:index]
 
 	var url string
-	protocol := "https"
+	protocol := httpsString
 	port := ""
 
 	if client.GetLedgerID().String() == "" {
-		protocol = "http"
+		protocol = httpString
 		port = ":5551"
 	}
 
--- a/contract_id.go
+++ b/contract_id.go
@@ -153,35 +153,35 @@ func (id ContractID) ToSolidityAddress() string {
 	return _IdToSolidityAddress(id.Shard, id.Realm, id.Contract)
 }
 
-// PopulateEvmAddress gets the actual `EvmAddress` field of the `ContractID` from the Mirror Node.
-func (id *ContractID) PopulateEvmAddress(client *Client) error {
-	result, err := id._MirrorNodeRequest(client, "evmAddress")
-	if err != nil {
-		return err
-	}
-
-	mirrorEvmAddress, ok := result["evm_address"].(string)
-	if !ok {
-		return errors.New("unexpected response format")
-	}
-
-	mirrorEvmAddress = strings.TrimPrefix(mirrorEvmAddress, "0x")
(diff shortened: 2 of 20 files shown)


contract_id.go
```diff
@@ -212,6 +212,47 @@ func (id ContractID) _ToProtobuf() *services.ContractID {
...
+	if index == -1 {
+		return nil, errors.New("invalid mirrorUrl format")
+	}
+	mirrorUrl = mirrorUrl[:index]
+
+	var url string
+	protocol := httpsString
+	port := ""
+
+	if client.GetLedgerID().String() == "" {
```
0xivanov, line 231: @ivaylonikolov7 fyi

mirror_node_contract_estimate_gas_query.go
```diff
@@ -0,0 +1,74 @@
...
+// This specifies the price of each unit of gas used in the transaction.
+func (query *MirrorNodeContractEstimateGasQuery) SetGasPrice(gasPrice int64) *MirrorNodeContractEstimateGasQuery {
+	query.setGasPrice(gasPrice)
+	return query
+}
+
+// Returns gas estimation for the EVM execution
+func (query *MirrorNodeContractEstimateGasQuery) Execute(client *Client) (uint64, error) {
+	return query.estimateGas(client)
+}
```
gsstoykov, line 74: Is there any difference between `MirrorNodeContractEstimateGasQuery` and `MirrorNodeContractCallQuery` if not for this `execute` method. You are housing the logic for `estimateGas` and the `call` in the parent which is `query`. Going down this path you would have to extend the logic in `query` while creating files and structs which just hold all the properties of the `query` and just adding an unnecessary layer of setters and getters. I would see the need of having `MirrorNodeContractEstimateGasQuery` and `MirrorNodeContractCallQuery` if they held at least some specific fields or functionalities.

gsstoykov (commented):
Few nits and needs a little design tweak. Otherwise LG.

mirror_node_contract_call_query.go
```diff
@@ -0,0 +1,81 @@
+package hiero
+
+// SPDX-License-Identifier: Apache-2.0
+
+// MirrorNodeContractCallQuery returns a result from EVM transient simulation of read-write operations.
+type MirrorNodeContractCallQuery struct {
```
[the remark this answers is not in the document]
0xivanov: The idea is that one query is for estimation and the other for call. Anyhow, we already agreed on the names with nana, simi and richard.

mirror_node_contract_estimate_gas_query.go
0xivanov: We are doing it solely for the naming. I initially had them be 1 query, which had `estimate` and `call` methods, but Simi and Nana wanted the query name to say what the query does - hence how we have 2 queries - `EstimateGasQuery` and `CallQuery`.

account_id.go
```diff
@@ -324,6 +324,9 @@ const (
 	EvmAddress
 )
 
+var httpString string = "http"
+var httpsString string = "https"
```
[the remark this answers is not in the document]
0xivanov: I might as well remove them, as they are not needed anymore i think

mirror_node_contract_query.go
```diff
@@ -0,0 +1,318 @@
...
+	// The gas limit
+	gasLimit *int64
+	// The gas price
+	gasPrice *int64
+	// The block number for the simulation
+	blockNumber *int64
+}
+
+// setContractID sets the contract instance to call
+func (query *mirrorNodeContractQuery) setContractID(contractID ContractID) {
```
gsstoykov, line 40: nit: Would also be good to differentiate between  the `query` naming here as we also have `query` in the transaction flow just for ease for code navigation and etc.

0xivanov: named `mirrorNodeContractQuery` for clarity

account_id.go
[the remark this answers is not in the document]
0xivanov: removed

gsstoykov (approved):
LGTM.

PR #1198, opened 2025-01-03
refactor: Move source go files in `/sdk` package

0xivanov:
### **Description**  
This PR reorganizes the project structure by introducing an `sdk` directory as the primary location for SDK-related functionality.  

#### **Changes Made**  
1. Created an `sdk` directory to serve as the central location for SDK source files.  
2. Moved all existing source files into the `sdk` directory.  

#### **Why Not Sub-Packages?**  
- Sub-packages are not introduced in this PR to keep the structure simple and minimize disruption to the current development workflow.  
- Sub-packaging will be considered in future iterations as the codebase grows, especially if functionality becomes more modular and requires logical grouping.  

#### **Why `/sdk` Instead of `/src`?**  
- Avoiding Non-Idiomatic Practices: Other popular Go libraries, such as `aws-sdk-go` and `go-redis`, avoid using `/src` and instead organize their files directly in meaningful directories like sdk, pkg, or at the root level.

Fixes https://github.com/hiero-ledger/hiero-sdk-go/issues/1205


.github/workflows/build.yml
```diff
@@ -48,10 +48,10 @@ jobs:
               run: go vet ./examples/...
 
             - name: Vet
-              run: go vet .
+              run: go vet ./sdk
```
gsstoykov, line 51: I think go vet should run on the whole project. Should go into examples/tck and so on if a need for another dir arises.

gsstoykov (commented):
I think that we shouldn't name the folder with the source `sdk` as the repo itself is the `sdk`.

0xivanov:
@gsstoykov maybe we can call it `hiero`, wdyt

.github/workflows/build.yml
0xivanov: For some reason `go vet ./...` does not scan /examples and /tck. Will investigate if we can scan every project with 1 command

0xivanov: TCK needs to be done in a separate command, because it's a different project.

gsstoykov (commented):
LGTM.

gsstoykov (approved):

PR #1235, opened 2025-01-27
feat: `TokenRejectFlow` refactor

0xivanov:
### Description
Currently, the `TokenRejectFlow` class and similar transaction flows are overly reliant on class hierarchies that treat flows as if they are transactions themselves. This design complicates the implementation and maintenance of transaction flows.  

This PR does not change the API, but only adds getters to allow modification to the individual transactions.

### DIFF
--- a/sdk/token_reject_flow.go
+++ b/sdk/token_reject_flow.go
@@ -3,17 +3,22 @@ package hiero
 // SPDX-License-Identifier: Apache-2.0
 
 type TokenRejectFlow struct {
-	ownerID           *AccountID
-	tokenIDs          []TokenID
-	nftIDs            []NftID
-	freezeWithClient  *Client
-	signPrivateKey    *PrivateKey
-	signPublicKey     *PublicKey
-	transactionSigner *TransactionSigner
+	TokenRejectTransaction     *TokenRejectTransaction
+	TokenDissociateTransaction *TokenDissociateTransaction
+	ownerID                    *AccountID
+	tokenIDs                   []TokenID
+	nftIDs                     []NftID
+	freezeWithClient           *Client
+	signPrivateKey             *PrivateKey
+	signPublicKey              *PublicKey
+	transactionSigner          *TransactionSigner
 }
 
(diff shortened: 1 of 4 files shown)


sdk/token_reject_flow.go
```diff
@@ -3,17 +3,22 @@ package hiero
...
-	freezeWithClient  *Client
-	signPrivateKey    *PrivateKey
-	signPublicKey     *PublicKey
-	transactionSigner *TransactionSigner
+	TokenRejectTransaction     *TokenRejectTransaction
+	TokenDissociateTransaction *TokenDissociateTransaction
+	ownerID                    *AccountID
+	tokenIDs                   []TokenID
+	nftIDs                     []NftID
+	freezeWithClient           *Client
```
gsstoykov, line 11: My initial idea was using `ownerIds`, `tokenIds` only on the transactions themselves. The freeze would also be done on the separate transactions. So the client won't be passed to the `TokenRejectFlow`. The only thing we need to use in the Flow would be a PrivateKey if sign is needed on that level.

sdk/token_reject_flow.go
```diff
@@ -93,14 +98,9 @@ func (tx *TokenRejectFlow) FreezeWith(client *Client) (*TokenRejectFlow, error)
 	return tx, nil
 }
 
-func (tx *TokenRejectFlow) _CreateTokenDissociateTransaction(client *Client) (*TokenDissociateTransaction, error) {
-	if client == nil {
-		return &TokenDissociateTransaction{}, nil
-	}
-	tokenDissociateTxn := NewTokenDissociateTransaction()
-
+func (tx *TokenRejectFlow) fillTokenDissociate() error {
```
gsstoykov, line 101: The same will be valid for the fillTransactions. As they could also be removed I think.

sdk/token_reject_flow.go
0xivanov: Yeah I faced some problems with keeping the API the same in JS using that method. I will check if it can work in Go.

Issue #1082, opened 2024-10-01
HIP-991: Permissionless revenue-generating Topic IDs for Topic Operators


0xivanov:
# Description

This HIP introduces new transactions and updates to existing transactions in the consensus node software. SDKs will need to support these new transactions and updates to existing ones so that developers can take advantage of the new features. Propose the API design all SDKs should adhere to.

### CustomFeeLimit

This message is used to specify the maximum custom fee that given user is willing to pay.

This class may be used in future transactions.

- `AccountId payerId` : A payer account identifier.
- `List<CustomFixedFee> customFees` : The maximum fees that the user is willing to pay.

### TopicCreateTransaction

- `Key feeScheduleKey` : The key which allows updates to the new topic’s fees.
    - **NOTE:** This value can be `null`.
    - `TopicCreateTransaction setFeeScheduleKey(Key)`
    - `Key getFeeScheduleKey()`
- `List<Key> feeExemptKeys` : The keys that will be exempt from paying fees.
    - **NOTE:** This value can be empty.
    - `TopicCreateTransaction setFeeExemptKeys(List<Key>)`
    - `TopicCreateTransaction addFeeExemptKey(Key)`
    - `TopicCreateTransaction clearFeeExemptKeys()`
    - `List<Key> getFeeExemptKeys()`
- `List<CustomFixedFee> customFees` : The fixed fees to assess when a message is submitted to the new topic.
    - **NOTE:** This value can be empty.
    - `TopicCreateTransaction setCustomFees(List<CustomFixedFee>)`
    - `TopicCreateTransaction addCustomFee(CustomFixedFee)`
    - `TopicCreateTransaction clearCustomFees()`
    - `List<CustomFixedFee> getCustomFees()`

### TopicUpdateTransaction

- `Key feeScheduleKey` : The desired new key which will allow for updates to the topic’s fees.
    - **NOTE:** This value can be `null` .
    - `TopicUpdateTransaction setFeeScheduleKey(Key)`
    - `TopicUpdateTransaction clearFeeScheduleKey()`
    - `Key getFeeScheduleKey()`
- `List<Key> feeExemptKeys` : The desired new keys that will be exempt from paying fees.
    - **NOTE:** This value can be empty.
    - `TopicUpdateTransaction setFeeExemptKeys(List<Key>)`
    - `TopicUpdateTransaction addFeeExemptKey(Key)`
    - `TopicUpdateTransaction clearFeeExemptKeys()`
[...]

PR #1250, opened 2025-02-07
feat(HIP-991): Permissionless revenue topics

0xivanov:
This HIP introduces new transactions and updates to existing transactions in the consensus node software. SDKs will need to support these new transactions and updates to existing ones so that developers can take advantage of the new features.

- [x] Implementation
- [x] Unit tests
- [x] E2E tests
- [x] Example

Fixes: #1082


sdk/topic_message_submit_transaction_unit_test.go
```diff
@@ -206,6 +206,43 @@ func TestUnitTopicMessageSubmitTransactionFreezeMock(t *testing.T) {
...
+
+	proto := transaction.build().GetConsensusSubmitMessage()
+	assert.Equal(t, proto.TopicID, topic._ToProtobuf())
+	assert.Equal(t, proto.Message, []byte("nothing to see here"))
+
+	// TODO: in theory should work, but `_BuildTransaction` sets only encoded `SignedTransactionBytes`
+	// baseTransactionProto, err := transaction._BuildTransaction(0)
+	// require.NoError(t, err)
+	// assert.Equal(t, baseTransactionProto.Body.MaxCustomFees[0].AccountId, client.GetOperatorAccountID()._ToProtobuf())
+	// assert.Equal(t, baseTransactionProto.Body.MaxCustomFees[0].Fees[0], customFixedFee._ToProtobuf())
```
gsstoykov, line 243: Maybe remove and just create an issue.

sdk/custom_fee_limit.go
```diff
@@ -0,0 +1,94 @@
+package hiero
+
+import "github.com/hiero-ledger/hiero-sdk-go/v2/proto/services"
+
+// SPDX-License-Identifier: Apache-2.0
+
+// A maximum custom fee that the user is willing to pay.
```
gsstoykov, line 7: note: Maybe will be appropriate to link HIPs when implementing classes like this.

sdk/topic_create_transaction.go
```diff
@@ -14,6 +14,9 @@ type TopicCreateTransaction struct {
 	autoRenewAccountID *AccountID
 	adminKey           Key
 	submitKey          Key
+	feeScheduleKey     Key
```
gsstoykov, line 17: Isn't Key defined as optional by the HIP. Would it be more appropriate to make it a pointer here?

gsstoykov (commented):
Overall LGTM. Just a few points.

sdk/topic_create_transaction_hip991_e2e_test.go
```diff
@@ -0,0 +1,866 @@
...
+
+	env.Client.SetOperator(env.OperatorID, env.OperatorKey)
+
+	// Verify the custom fee charged
+	accountInfo, err := NewAccountInfoQuery().
+		SetAccountID(payerId).
+		Execute(env.Client)
+	require.NoError(t, err)
+
+	assert.True(t, accountInfo.Balance.AsTinybar() < hbar/2)
```
gsstoykov, line 293: nit: If balance is `hbar` and then a fee is applied for `hbar/2` this assertion would be intuitively wrong as `(hbar/2 < hbar/2) == false` I guess this only is valid because of side fees charged from the account. would just verify like `beforeHbar` < `afterHbar`

sdk/topic_create_transaction_hip991_e2e_test.go
```diff
@@ -0,0 +1,866 @@
...
+
+	_, err = resp.SetValidateStatus(true).GetReceipt(env.Client)
+	require.NoError(t, err)
+
+	customFeeLimit := NewCustomFeeLimit().
+		SetPayerId(payerId).
+		AddCustomFee(NewCustomFixedFee().SetAmount(2).SetDenominatingTokenID(tokenId))
+
+	// Submit a message to the revenue generating topic with custom fee limit
+	env.Client.SetOperator(payerId, payerPrivateKey)
```
gsstoykov, line 394: Why do we set the `Client` operator here. Could we just not just freeze -> sign with payer account?

gsstoykov: This is valid for the most of the PR.

sdk/topic_create_transaction_hip991_e2e_test.go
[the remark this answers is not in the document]
0xivanov: yeah i wanted to make them separate, will name them something like `revenue-generating_topic...`

sdk/topic_message_submit_transaction_unit_test.go
0xivanov: yeah, totally forgot about this :D

sdk/topic_create_transaction.go
0xivanov: the Key is an interface which by default is a pointer, i'm pretty sure. It works something like an array - it's default value is nil.

sdk/topic_create_transaction_hip991_e2e_test.go
0xivanov: Yeah i get the point.
Can you elaborate on the `beforeHbar < afterHbar` statement? I'm confused since the hbar balance before should be greater then hbar after.

If you meant `beforeHbar > afterHbar` - verifying that the balance is lower after applying the custom fees, that would be true even without the custom fees, because of the transaction fees for the message submit.

sdk/topic_create_transaction_hip991_e2e_test.go
0xivanov: I do it in order to make the payer a payer for the txn. Only signing won't do it. This is one way - set it as an operator. The other way is to `TransactionIDGenerate` -> freeze -> sign, which is more verbose with Go's error handling and stuff.

sdk/topic_create_transaction_hip991_e2e_test.go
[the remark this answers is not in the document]
0xivanov: Done

sdk/topic_create_transaction_hip991_e2e_test.go
```diff
@@ -0,0 +1,866 @@
...
+	exemptKey7, err := PrivateKeyGenerateEd25519()
+	require.NoError(t, err)
+	exemptKey8, err := PrivateKeyGenerateEd25519()
+	require.NoError(t, err)
+	exemptKey9, err := PrivateKeyGenerateEd25519()
+	require.NoError(t, err)
+	exemptKey10, err := PrivateKeyGenerateEd25519()
+	require.NoError(t, err)
+	exemptKey11, err := PrivateKeyGenerateEd25519()
+	require.NoError(t, err)
```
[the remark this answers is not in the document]
0xivanov: Done

gsstoykov (approved):
LGTM.

PR #1285, opened 2025-03-04
fixed ecdsa verify method

zzspoon:
**Description**:
Ecdsa pubkey fails to verify messages.
Should verify the sha3 hash instead of the raw message bytes

**Example**:
```go
func main() {
    message := []byte("message")

    k, _ := hiero.PrivateKeyGenerateEcdsa()
    pub := k.PublicKey()

    sig := k.Sign(message)
    ver := pub.Verify(message, sig)

    fmt.Println(ver) // true
}
```
**Checklist**


zzspoon:
This is my first pull request (for anything, ever...) so my apologies if something is not proper, but I found this bug and it's quite a simple fix

SimiHunjan:
Someone from the team will review and get back to you! Appreciate your first contribution!!

0xivanov:
@zzspoon will take a look

0xivanov:
@zzspoon Currently we are expecting the user to pass the keccak hashed message as an argument. I need some time to check the behaviour in the other SDKs, because they need to be aligned.

zzspoon:
@0xivanov Thank you! This does appear to be the behavior of the [Java SDK](https://github.com/hiero-ledger/hiero-sdk-java/blob/main/sdk/src/main/java/com/hedera/hashgraph/sdk/PublicKeyECDSA.java#L69)

We can also see that the test uses pubkey.verify on the [message itself](https://github.com/hiero-ledger/hiero-sdk-java/blob/53c7c1607d6cbd3b5b4f0a544b932f28d2078adb/sdk/src/test/java/com/hedera/hashgraph/sdk/KeyTest.java#L42).

My change to the test tries to keep the old behavior as well, but perhaps this is not necessary...

zzspoon:
also, the [calcKeccak256](https://github.com/hiero-ledger/hiero-sdk-java/blob/main/sdk/src/main/java/com/hedera/hashgraph/sdk/Crypto.java#L236) function of the Java SDK is not public. My proposal would allow encapsulation of the [Keccak256Hash](https://github.com/hiero-ledger/hiero-sdk-go/blob/32cce664f079d2e854fea1f8b9712e79cc319f44/sdk/crypto.go#L900) and the [Hash type](https://github.com/hiero-ledger/hiero-sdk-go/blob/32cce664f079d2e854fea1f8b9712e79cc319f44/sdk/crypto.go#L926). I have not yet pushed the unexported Keccak256Hash func and Hash type

0xivanov:
@zzspoon Hashing the message inside the function looks like the behaviour in other SDKs. 
Although I am afraid changing it in go might cause breaking change. I am thinking how we can  prevent that from happening.

zzspoon:
I understand. My intuition was that if someone wanted to use Verify, they would have done basically what I did. Anyone using an arbitrary keytype would first need to determine the keytype prior to using Verify, since the behavior is different with ed25519. Because the method is not keytype-agnostic, I considered it a bug

0xivanov:
Would you be able to create a new fixed method and deprecate the old one.
There will be some code duplication but it's fine for me.

zzspoon:
@0xivanov I was thinking about a possibly elegant solution: Remove [Hash.Bytes](https://github.com/hiero-ledger/hiero-sdk-go/blob/32cce664f079d2e854fea1f8b9712e79cc319f44/sdk/crypto.go#L928) and replace it with Hash.BytesRaw and Hash.BytesVerify. 

BytesRaw would behave as the normal hash func, while BytesVerify would append the hash to a unique prefix. PublicKey.Verify may then parse the message and check for the prefix. If found, verify message[len(prefix):] directly. Otherwise, hash the message and verify BytesRaw.

Anyone compiling the new version while using Hash.Bytes would see the breakage, however the functionality would be identical and requires a plain search-and-replace. Change hash.Bytes to hash.BytesVerify and it behaves as expected, while everyone else may pass a message directly to PublicKey.Verify, regardless of key type

Not sure if this is adequate or desirable

0xivanov:
I don't like the idea to just remove something. Yeah I get that it would be easy find + replace method, but we've had cases where users would complain if we remove or change stuff without deprecating notice. We could get in trouble.

0xivanov:
@zzspoon Nice work 🚀

sdk/ed25519_public_key.go
```diff
@@ -161,7 +161,7 @@ func (pk _Ed25519PublicKey) _VerifyTransaction(tx *Transaction[TransactionInterf
 		for _, sigPair := range tx.SigMap.GetSigPair() {
 			if bytes.Equal(sigPair.GetPubKeyPrefix(), pk._Bytes()) {
 				found = true
-				if !pk._Verify(tx.BodyBytes, sigPair.GetEd25519()) {
+				if !pk._VerifySignedMessage(tx.BodyBytes, sigPair.GetEd25519()) {
```
0xivanov, line 164: Can you please create an integration test that calls `VerifyTransaction` on a ECDSA key. You can look at `TestIntegrationAccountDeleteTransactionCanExecute` and do the same test flow but with ECDSA key. 
It seems like we don't have coverage for this.

zzspoon:
Found a good home for test where both keytypes live together

Some [methods](https://github.com/hiero-ledger/hiero-sdk-go/blob/main/sdk/crypto.go#L407) were marked as deprecated, however they were (purposely?) dodging the linter, causing tck to complain. The corresponding methods in the [js](https://github.com/hiero-ledger/hiero-sdk-js/blob/4af517c8277784d77afc85775bccfb7d42cd1149/src/PublicKey.js#L66) and [java](https://github.com/hiero-ledger/hiero-sdk-java/blob/b1579fd4cc2e9de364f5d1949c98c24dd3ec1916/sdk/src/main/java/com/hedera/hashgraph/sdk/PublicKey.java#L77) sdk do not seem to be deprecated, so I mark them as "deprecated pending review". 

Apologies for bloated pr, I should have opened a separate one for the docs

0xivanov:
Not sure why the test is failing :D, investigating now.

PR #1318, opened 2025-03-24
chore: add codecov

0xivanov:
**Description**:
Removes usages of `client.GetNetworkName()` since it's deprecated and adds unit tests.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:
Since `GetLedgerID` never returned `nil` codecov complained. I change it to return `nil` when the client is created for custom network (as it's in Java SDK) and eliminated the null pointers.

**Checklist**


sdk/client.go
```diff
@@ -537,7 +536,11 @@ func (client *Client) SetNetworkName(name NetworkName) {
 
 // Deprecated: Use GetLedgerID instead
 func (client *Client) GetNetworkName() *NetworkName {
-	name, _ := client.GetLedgerID().ToNetworkName()
+	ledgerID := client.GetLedgerID()
+	var name NetworkName
+	if ledgerID != nil {
+		name, _ = ledgerID.ToNetworkName()
```
gsstoykov, line 542: nit: Shouldn't we check ToNetworkName err case if it returns one?

sdk/client.go
```diff
@@ -537,7 +536,11 @@ func (client *Client) SetNetworkName(name NetworkName) {
...
+	}
 	return &name
```
gsstoykov, line 544: nit: It's better if we explicitly return `nil` when the logic leads to it. Return &name otherwise.

sdk/client.go
0xivanov: `GetNetworkName` does not return error so we cannot propagate it.

sdk/client.go
0xivanov: You mean to return nil if `ledgerID == nil`?
Here `var name NetworkName`'s value is nil by default so `return &name` would return nil.

gsstoykov (approved):

PR #1324, opened 2025-03-26
feat: Аdd `GetRecoveryId` and align ECDSA signing with other SDKs

### LINKED ISSUE #909: Add recovery id calculation when signing ethereum call transactions
### Problem

Currently there is no way to calculate recovery id (v) when we sign ethereum transactions from go SDK.

### Solution

The recovery ID is an index from 0 to 3, which indicates which of the 4 possible public keys is the correct one. Because the key recovery operation yields multiple potential public keys, the correct key must either be stored alongside the signature or you must be willing to try each recovery ID in turn until you find one that outputs the key you are expecting.

Recovery operation is described in [SEC1v2, section 4.1.6](https://www.secg.org/sec1-v2.pdf).

### Alternatives

_No response_

0xivanov:
**Description**:

This PR introduces two key enhancements to the Hedera Go SDK's handling of ECDSA private keys:

1. **Addition of `GetRecoveryId` Method**: A new method, `GetRecoverID`, has been added to the `PrivateKey` struct for ECDSA keys. This method retrieves the recovery ID (also known as the 'v' value) associated with ECDSA signatures, facilitating signature verification processes.
2. **Modification of the `Sign` Method Output**: The `Sign` method for ECDSA private keys has been updated to return only the `r` and `s` components of the signature, reducing the output from 65 bytes to 64 bytes. This change aligns the SDK's behavior with standard ECDSA signature formats, which typically include only the `r` and `s` values.

These updates enhance the SDK's compatibility with standard ECDSA operations and improve interoperability with other SDKs.

### Usage

```go
	eip1559tx := &RLPItem{}
	eip1559tx.AssignList()
	eip1559tx.PushBack(NewRLPItem(VALUE_TYPE).AssignValue(chainId))
        // fill other required fields
        ...
	messageBytes, err := eip1559tx.Write()
	if err != nil {
		return nil, err
	}
	messageBytes = append([]byte{0x02}, messageBytes...)

	sig := ecdsaPrivateKey.Sign(messageBytes)

	r := sig[0:32]
	s := sig[32:64]
	v := ecdsaPrivateKey.GetRecoveryId(r, s, messageBytes)

	eip1559tx.PushBack(NewRLPItem(VALUE_TYPE).AssignValue(recIdBytes))
	eip1559tx.PushBack(NewRLPItem(VALUE_TYPE).AssignValue(r))
	eip1559tx.PushBack(NewRLPItem(VALUE_TYPE).AssignValue(s))
```

### Deprecation notice ⚠️ 
`VerifySignature` no longer maintained, since it requires the full 65 byte signature and pre-hashing using `Keccak256Hash`. `PublicKey.VerifySignedMessage` is preferred.

## Migration guide
```go
// before
	hash := Keccak256Hash([]byte("aaa"))
	sig := key.Sign([]byte("aaa"))
	isValid := VerifySignature(key.PublicKey().BytesRaw(), hash.Bytes(), sig)

// after
        sig := key.Sign([]byte("aaa"))
	isValid := key.PublicKey().VerifySignedMessage([]byte("aaa"), sig)
```

**Related issue(s)**:

Fixes #909

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/sdk/ecdsa_private_key.go
+++ b/sdk/ecdsa_private_key.go
@@ -209,22 +209,17 @@ func _ECDSAPrivateKeyReadPem(source io.Reader, passphrase string) (*_ECDSAPrivat
 
 func (sk _ECDSAPrivateKey) _Sign(message []byte) []byte {
 	hash := Keccak256Hash(message)
-	signature := ecdsa.Sign(sk.keyData, hash.Bytes())
-
-	r := signature.R()
-	rBytes := r.Bytes()
-	s := signature.S()
-	sBytes := s.Bytes()
-
-	return append(rBytes[:], sBytes[:]...)
+	return ecdsa.SignCompact(sk.keyData, hash.Bytes(), true)[1:]
 }
 
 func (sk _ECDSAPrivateKey) getRecoveryId(r []byte, s []byte, message []byte) int {
 	hash := Keccak256Hash(message)
 
-	// nolint
-	sig := append(r, s...)
+	// Merge r and s into a single byte slice
+	r = append(r, s...)
+	sig := r
 
+	// Try to recover the public key using the signature and the hash
 	for i := 0; i < 4; i++ {
 		sigWithRecID := make([]byte, 65)
 		copy(sigWithRecID[1:], sig)
--- a/sdk/ecdsa_private_key.go
+++ b/sdk/ecdsa_private_key.go
@@ -209,7 +209,14 @@ func _ECDSAPrivateKeyReadPem(source io.Reader, passphrase string) (*_ECDSAPrivat
 
 func (sk _ECDSAPrivateKey) _Sign(message []byte) []byte {
 	hash := Keccak256Hash(message)
-	return ecdsa.SignCompact(sk.keyData, hash.Bytes(), true)[1:]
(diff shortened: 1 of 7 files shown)


sdk/ecdsa_private_key.go
```diff
@@ -209,7 +209,39 @@ func _ECDSAPrivateKeyReadPem(source io.Reader, passphrase string) (*_ECDSAPrivat
...
+
+	r := signature.R()
+	rBytes := r.Bytes()
+	s := signature.S()
+	sBytes := s.Bytes()
+
+	return append(rBytes[:], sBytes[:]...)
+}
+
+func (sk _ECDSAPrivateKey) getRecoveryId(r []byte, s []byte, message []byte) int {
```
gsstoykov, line 222: Wouldn't it be better if we have `sig` passed directly here it would also remove the need for the `// nolint` in the function.

0xivanov: Yeah that's a valid point but we've already implemented it like this in the Java SDK. Maybe it's nice to have consistency and i guess there could be some interoperability in the future if we merge them into 1 byte array.
I will look into the lint issue though.

gsstoykov (approved):

PR #1345, opened 2025-04-09
fix: `INVALID_NODE_ACCOUNT` when setting nodes for paid queries 

0xivanov:
**Description**:
TBD

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/sdk/query.go
+++ b/sdk/query.go
@@ -101,10 +101,6 @@ func (q *Query) getCost(client *Client, e QueryInterface) (Hbar, error) {
 
 	q.pb = e.buildQuery()
 
-	if q.isPaymentRequired && len(q.paymentTransactions) > 0 {
-		q.paymentTransactionIDs._Advance()
-	}
-
 	q.pbHeader.ResponseType = services.ResponseType_COST_ANSWER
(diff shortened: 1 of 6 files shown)


sdk/query.go
```diff
@@ -101,10 +101,6 @@ func (q *Query) getCost(client *Client, e QueryInterface) (Hbar, error) {
 
 	q.pb = e.buildQuery()
 
-	if q.isPaymentRequired && len(q.paymentTransactions) > 0 {
-		q.paymentTransactionIDs._Advance()
```
0xivanov, line 105: I think this snippet is useless because:
1. `paymentTransactionIDs` are not used in the internal logic.
2. We create new slice on line 97, so think it's impossible `len(q.paymentTransactions) > 0` to be true here.

0xivanov:
There are some tests failing, but it's expected. For example - ethereum transaction's chain id, long term scheduled txns are not yet enabled, and cannot update file 111 on testnet

Here is the output of the test run:
[output.txt](https://github.com/user-attachments/files/19683158/output.txt)

gsstoykov (approved):

PR #1347, opened 2025-04-10
feat(HIP-551): `BatchTransaction`

### LINKED ISSUE #1270: HIP-551: Batch Transactions
## Description

HIP: https://hips.hedera.com/hip/hip-551

This HIP introduces new transactions and updates to existing transactions in the consensus node software. SDKs will need to support these new transactions and updates to existing ones so that developers can take advantage of the new features. Propose the API design all SDKs should adhere to.

0xivanov:
**Description**:
This HIP defines a mechanism to execute batch transactions such that a series of transactions (HAPI calls) depending on each other can be rolled into one transaction that passes the ACID test (atomicity, consistency, isolation, and durability).
Example usage can be found in 
`examples/batch-transaction/main.go`

### BatchTransaction

- `List<Transaction> innerTransactions`: The list of `Transaction`s to execute as part of the batch.
    - `BatchTransaction setInnerTransactions(List<Transaction>)`
    - `BatchTransaction addInnerTransaction(Transaction)`
    - `List<Transaction> getInnerTransactions()`
- `List<TransactionId> getInnerTransactionIds`: List of innerTransactionIds that have been added to the batch. Later used to query them and verify their status.

### Transaction

- `Key batchKey`: The key that must sign the `BatchTransaction` of which this `Transaction` is a part.
    - `Transaction batchify(Client client, Key batchKey)` : this method sets the batch key and marks a transaction as part of a batch transaction (inner transaction). The transaction is signed by the client of the operator and frozen.
    - `Transaction setBatchKey(Key)`: sets the batch key
    - `Key getBatchKey()` : gets the batch key
- `freezeWith`: Additional check if batchKey is `null`.
    - If not present → indication that this transaction is not meant to be inside a batch
    - If preset → indication that this transaction is part of a batch and setting `nodeAccountId` to `0.0.0`

**Related issue(s)**:

Fixes #1270

**Notes for reviewer**:

**Checklist**


0xivanov:
Tested against testnet everything looks good.

sdk/batch_transaction.go
```diff
@@ -0,0 +1,248 @@
...
+// - It is frozen
+// - It has a batch key set
+func (tx *BatchTransaction) validateInnerTransaction(innerTransaction TransactionInterface) error {
+	// Check for nil transaction
+	if innerTransaction == nil {
+		return errInnerTransactionNil
+	}
+
+	// Validate transaction type is not blacklisted
+	txType := reflect.TypeOf(innerTransaction)
```
gsstoykov, line 96: If we have something like 
```
type BlacklistedTransaction interface {
	isBlacklisted()
}
``` 
We would remove the need to reflect and the code will be a lot faster and lightweight memory wise.

sdk/batch_transaction.go
```diff
@@ -0,0 +1,248 @@
...
+//
+// Returns:
+// - *BatchTransaction: The BatchTransaction instance for fluent API calls
+// - error: If any validation fails
+func (tx *BatchTransaction) SetInnerTransactions(transactions []TransactionInterface) *BatchTransaction {
+	tx._RequireNotFrozen()
+
+	// Validate all transactions before setting
+	for _, transaction := range transactions {
+		if err := tx.validateInnerTransaction(transaction); err != nil {
```
gsstoykov, line 130: We usually don't implement a validation layer of the SDK inputs. Should we aim to have such a layer in the future?

sdk/batch_transaction.go
```diff
@@ -0,0 +1,248 @@
...
+
+func (tx BatchTransaction) buildScheduled() (*services.SchedulableTransactionBody, error) {
+	return nil, errors.New("cannot schedule `BatchTransaction`")
+}
+
+func (tx BatchTransaction) buildProtoBody() *services.AtomicBatchTransactionBody {
+	body := &services.AtomicBatchTransactionBody{}
+	for _, innerTransaction := range tx.innerTransactions {
+		request := innerTransaction.makeRequest()
+		switch request := request.(type) {
```
gsstoykov, line 224: nit: Isn't this switch redundant.

gsstoykov: Could be something like:
```
	transaction, ok := request.(*services.Transaction)
		if !ok {
			return nil, fmt.Errorf("invalid inner transaction type: %T", request)
		}
```

sdk/batch_transaction.go
0xivanov: So you suggest to add `isBlacklisted()` to the transaction interface?
I'd have to implement it in every transaction.

sdk/batch_transaction.go
0xivanov: If the transaction is as complex as the atomic batch, where the user has to comply with a lot of constraints - it'd say yes

sdk/transaction.go
```diff
@@ -142,81 +143,97 @@ func TransactionFromBytes(data []byte) (TransactionInterface, error) { // nolint
...
-		if len(transactionFromList.SignedTransactionBytes) == 0 {
-			txIsSigned = false
-			if err := protobuf.Unmarshal(transactionFromList.BodyBytes, &body); err != nil { // nolint
-				return nil, errors.Wrap(err, "error deserializing BodyBytes in TransactionFromBytes")
-			}
-		} else { // If the transaction is signed/locked
-			if err := protobuf.Unmarshal(transactionFromList.SignedTransactionBytes, &signedTransaction); err != nil {
-				return nil, errors.Wrap(err, "error deserializing SignedTransactionBytes in TransactionFromBytes")
-			}
+	if len(list.TransactionList) == 0 {
```
0xivanov, line 146: note: had to add handling for single transaction decoding.

sdk/batch_transaction.go
0xivanov: Done

sdk/batch_transaction.go
gsstoykov: Maybe there is no need for now. If a more general use case for restricting transactions appear we could then implement such an interface.

gsstoykov (approved):

PR #1348, opened 2025-04-10
feat: Extend `SetKeyWithAlias` funcs to support `PublicKey`

### LINKED ISSUE #1357: Improve setECDSAKeyWithAlias and setKeyWithAlias to Accept Public or Private ECDSA Keys
Current SDK methods related to key setting during transaction/account creation — specifically `setECDSAKeyWithAlias` and `setKeyWithAlias` — have inconsistencies in expected parameter types, making them less intuitive and leading to integration friction.

---

###  Problem

- `.setECDSAKeyWithAlias()` expects a `PrivateKey`
- `.setKeyWithAlias()` expects both a `Key` and a `PrivateKey`
- `.setKeyWithoutAlias()` expects a `Key`

These inconsistencies are especially problematic when only the `PublicKey` is available on the frontend (as it should be — `PrivateKey` should **not** be transmitted off-device for security reasons).

---

### Suggested Solution

Update and unify the method signatures to accept a more flexible input type, such as `Key`, which can represent either an ECDSA `PrivateKey` or a `PublicKey`. The updated behavior should support both scenarios:

- If a **PrivateKey** is provided: use it to set the key and internally derive the alias from its corresponding **public key**.
- If a **PublicKey** is provided: use it directly to set the key and derive the alias from that **public key**.

This ensures that both `PrivateKey` and `PublicKey` inputs result in the same outcome:
- The key is set on the transaction
- The alias is derived from the ECDSA public key

By allowing both types, the SDK would support secure frontend-backend workflows where only the public key is transmitted from the client, without breaking functionality or requiring workarounds.

---

### Workaround

- A temporary workaround using `.setKeyWithoutAlias(...).setAlias(...)` was suggested, but feels non-intuitive for developers new to the SDK.

0xivanov:
**Description**:
Update and unify the method signatures to accept a more flexible input type, such as `Key`, which can represent either an ECDSA `PrivateKey` or a `PublicKey`. The updated behavior should support both scenarios:

- If a **PrivateKey** is provided: use it to set the key and internally derive the alias from its corresponding **public key**.
- If a **PublicKey** is provided: use it directly to set the key and derive the alias from that **public key**.

This ensures that both `PrivateKey` and `PublicKey` inputs result in the same outcome:
- The key is set on the transaction
- The alias is derived from the ECDSA public key

By allowing both types, the SDK would support secure frontend-backend workflows where only the public key is transmitted from the client, without breaking functionality or requiring workarounds.

**Related issue(s)**:

Fixes #1357

**Notes for reviewer**:

**Checklist**


ivaylogarnev-limechain (dismissed):
LGTM

gsstoykov (approved):
LGTM.

PR #1391, opened 2025-05-30
chore: remove-embeded-realms-and-shards

0xivanov:
**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/sdk/account_id.go
+++ b/sdk/account_id.go
@@ -98,6 +98,8 @@ func AccountIDFromEvmPublicAddress(s string) (AccountID, error) {
 
 // AccountIDFromSolidityAddress constructs an AccountID from a string
 // representation of a _Solidity address
+// "000000000000000000000000000000000000138D" -> 0.0.59 ?
+// "302a300506032b6570032100114e6abc371b82da" -> 0.0.302a300506032b6570032100114e6abc371b82da ?
 // Deprecated
 func AccountIDFromSolidityAddress(s string) (AccountID, error) {
 	evmAddress, err := hex.DecodeString(s)
--- a/sdk/account_id.go
+++ b/sdk/account_id.go
@@ -98,17 +98,18 @@ func AccountIDFromEvmPublicAddress(s string) (AccountID, error) {
 
 // AccountIDFromSolidityAddress constructs an AccountID from a string
 // representation of a _Solidity address
+// Deprecated
 func AccountIDFromSolidityAddress(s string) (AccountID, error) {
-	shard, realm, account, err := _IdFromSolidityAddress(s)
+	evmAddress, err := hex.DecodeString(s)
 	if err != nil {
 		return AccountID{}, err
 	}
-
 	return AccountID{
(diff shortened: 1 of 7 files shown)


sdk/account_id.go
```diff
@@ -98,17 +98,20 @@ func AccountIDFromEvmPublicAddress(s string) (AccountID, error) {
 
 // AccountIDFromSolidityAddress constructs an AccountID from a string
 // representation of a _Solidity address
+// "000000000000000000000000000000000000138D" -> 0.0.59 ?
+// "302a300506032b6570032100114e6abc371b82da" -> 0.0.302a300506032b6570032100114e6abc371b82da ?
```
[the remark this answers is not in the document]
SimiHunjan: yes we do not want to encode them in the 20 byte evm address.

If the user wants to create an entity ID with the evm address it would have to result in 0.0.evmAddress.

sdk/account_id.go
```diff
@@ -98,17 +98,20 @@ func AccountIDFromEvmPublicAddress(s string) (AccountID, error) {
...
-		Realm:    realm,
-		Account:  account,
-		checksum: nil,
+		Shard:           0,
+		Realm:           0,
+		Account:         0,
+		checksum:        nil,
+		AliasEvmAddress: &evmAddress,
 	}, nil
 }
```
SimiHunjan, line 116: we would not change the return on this. Rather mark it as deprecated and if someone needed to create an entity ID out of the EVM address there should be either an existing method or new method created to provide such functionality.

sdk/contract_id.go
```diff
@@ -107,15 +107,16 @@ func ContractIDFromEvmAddress(shard uint64, realm uint64, evmAddress string) (Co
 // Does not populate ContractID.EvmAddress
 // Deprecated
 func ContractIDFromSolidityAddress(s string) (ContractID, error) {
```
SimiHunjan, line 109: This is already marked as deprecated so I don't think we need to do anything here

sdk/topic_id_unit_test.go
```diff
@@ -65,3 +65,58 @@ func TestUnitTopicIDToSolidityAddress(t *testing.T) {
...
+
+	// Test with an EVM address that's too short
+	_, err := TopicIDFromEvmAddress(0, 0, "abc123")
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "EVM address is not a correct long zero addres")
+
+	// Test with an EVM address that's too long
+	_, err = TopicIDFromEvmAddress(0, 0, "0123456789abcdef0123456789abcdef0123456789abcdef")
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "EVM address is not a correct long zero addres")
```
gsstoykov, line 80: issue: I think that at some point we should create a way to move out of `require.Contains` as it is easy to break and hard to maintain. Could move to defined Error types for the SDK.

PR #1395, opened 2025-06-03
WIP: Init client with shard and realm

0xivanov:
**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/sdk/client.go
+++ b/sdk/client.go
@@ -34,6 +34,8 @@ type Client struct {
 	operator *_Operator
 
 	network                         _Network
+	shard                           uint64
+	realm                           uint64
 	mirrorNetwork                   *_MirrorNetwork
 	autoValidateChecksums           bool
 	defaultRegenerateTransactionIDs bool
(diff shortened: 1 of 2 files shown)


sdk/client.go
```diff
@@ -85,7 +87,15 @@ func ClientForMirrorNetworkWithRealmAndShard(mirrorNetwork []string, realm uint6
...
 func ClientForNetwork(network map[string]AccountID) *Client {
 	net := _NewNetwork()
-	client := _NewClient(net, []string{}, nil, true)
+	client := _NewClient(net, []string{}, nil, true, 0, 0)
+	_ = client.SetNetwork(network)
+	return client
+}
+
+// ClientForNetwork constructs a client given a set of nodes.
+func ClientForNetworkShardRealm(network map[string]AccountID, shard uint64, realm uint64) *Client {
```
0xivanov, line 96: Naming is up for discussion

sdk/client.go
```diff
@@ -240,12 +286,20 @@ func ClientFromConfigWithoutScheduleNetworkUpdate(jsonBytes []byte) (*Client, er
...
 	if err != nil {
 		return nil, err
 	}
 
+	if clientConfig.Shard != 0 {
+		shard = clientConfig.Shard
+	}
+	if clientConfig.Realm != 0 {
+		realm = clientConfig.Realm
+	}
```
gsstoykov, line 302: if `clientConfig.Shard == 0` the default value for the temp vars shard, realm gets used which is zero but the clientConfig.Shard is already zero by default so in both cases:
`client = _NewClient(network, arr, nil, shouldScheduleNetworkUpdate, shard, realm)` seems equivalent to `client = _NewClient(network, arr, nil, shouldScheduleNetworkUpdate, clientConfig.Shard, clientConfig.Realm)` so the temp vars could be removed together with those checks.

gsstoykov (approved):
LGTM

PR #1500, opened 2025-09-16
chore: Update file permissions to remove x permissions

jwagantall:


rbarker-dev (changes requested):
This will break pre-commit and install hooks.

PR #1715, opened 2026-05-14
fix: stabilize node readmit selection

### LINKED ISSUE #1714: TestUnitConcurrentGetNodeReadmit can panic locally on Windows
## Summary

`TestUnitConcurrentGetNodeReadmit` can panic locally on Windows with:

```text
panic: failed to find a healthy working node
```

This was observed while validating #1708, but it also reproduces on a clean `origin/main` checkout, so it does not appear to be introduced by that PR.

## Environment

```text
go version go1.25.7 windows/amd64
```

Baseline reproduced on clean `origin/main` at:

```text
d91f3f2b1938fcd651b43bff9eeed7ca33fe5ec1
```

## Reproduction

```bash
go test ./sdk -tags="unit" -v -run TestUnitConcurrentGetNodeReadmit -count=10 -timeout 9999s
```

## Observed

The test can panic with:

```text
panic: failed to find a healthy working node
sdk/managed_network.go:185
sdk/network_unit_test.go:85
```

## Notes

The PR-specific test from #1708 passes locally:

```bash
go test ./sdk -tags="unit" -v -run TestUnitHbarAllowance -timeout 9999s
```

The GitHub Actions `Unit and Integration Tests (unit)` job is also passing on #1708, so this may be environment-specific or timing-sensitive.

peter941221:
## Description

Fixes #1714.

`_GetNode()` previously called `_ReadmitNodes()` and then released the write lock before taking a read lock to inspect `healthyNodes`. Under concurrent `_GetNode()` / `_IncreaseBackoff()` calls, another goroutine could remove all healthy nodes in that gap, causing the existing `TestUnitConcurrentGetNodeReadmit` test to panic locally on Windows.

This keeps node readmission and node selection in the same critical section, and treats readmit times equal to `now` as due. The panic behavior is preserved when there is genuinely no healthy or due-for-readmit node.

## Testing

- `go test ./sdk -tags="unit" -v -run TestUnitConcurrentGetNodeReadmit -count=50 -timeout 9999s`
- `go test ./sdk -tags="unit" -v -run TestUnitConcurrentNodeAccess -count=10 -timeout 9999s`
- `go test ./sdk -tags="unit" -v -run TestUnitConcurrentNodeGetChannel -count=10 -timeout 9999s`
- `go test ./sdk -tags="unit" -v -run TestUnitGetNodePanicNoHealthyNodes -count=10 -timeout 9999s`
- `go test ./sdk -tags="unit" -v -run TestUnitReadmitNodes -count=10 -timeout 9999s`

Notes:
- `go test ./sdk -tags="unit" -timeout 9999s` currently fails locally with an unrelated `grpc: the server has been stopped` panic from `sdk/mock_test.go`.
- `go test -race ...` could not run in this Windows environment because `runtime/cgo` requires `gcc`, which is not available in PATH.

### DIFF
--- a/sdk/managed_network.go
+++ b/sdk/managed_network.go
@@ -100,7 +100,11 @@ func (mn *_ManagedNetwork) _ReadmitNodes() {
 	mn.healthyNodesMutex.Lock()
 	defer mn.healthyNodesMutex.Unlock()
 
-	if mn.earliestReadmitTime.Before(now) {
+	mn._ReadmitNodesLocked(now)
+}
+
+func (mn *_ManagedNetwork) _ReadmitNodesLocked(now time.Time) {
+	if !mn.earliestReadmitTime.After(now) {
 		nextEarliestReadmitTime := now.Add(mn.maxNodeReadmitPeriod)
 
 		for _, node := range mn.nodes {
@@ -122,7 +126,7 @@ func (mn *_ManagedNetwork) _ReadmitNodes() {
 				}
 			}
 
-			if node._GetReadmitTime().Before(now) {
+			if !node._GetReadmitTime().After(now) {
 				mn.healthyNodes = append(mn.healthyNodes, node)
 			}
 		}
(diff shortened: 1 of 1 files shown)


peter941221:
Locking note for review:

`_GetNode()` now takes the `healthyNodesMutex` write lock because the readmission step and the healthy-node snapshot need to be atomic with respect to `_IncreaseBackoff()`.

Before this change, `_GetNode()` called `_ReadmitNodes()`, released that write lock, and only then took an `RLock` to inspect `healthyNodes`. In the gap, concurrent `_IncreaseBackoff()` calls could remove the recently readmitted nodes and leave `healthyNodes` empty, which is the panic reported in #1714.

The lock is not held while selecting a random node. The method clones `healthyNodes` while protected, releases the lock, then selects from the snapshot. That keeps the critical section small while preventing the empty-list race.

peter941221:
One implementation note for review: this PR intentionally includes two small pieces in the same file because they address the same failing path.

```text
_GetNode()
  -> readmit nodes
  -> inspect/select healthyNodes
```

The locking change keeps readmission and the healthy-node snapshot in one critical section, so concurrent `_IncreaseBackoff()` calls cannot drain `healthyNodes` between readmission and selection.

The `Before(now)` to `!After(now)` change handles the zero-duration readmit case used by `TestUnitConcurrentGetNodeReadmit`: if a node's readmit time is exactly `now`, it should be considered due for readmission. This matters more on Windows where timer granularity can make the equality boundary easier to hit.

If you would prefer, I can split the equality-boundary change from the locking change, but I kept them together here because both are part of making the same readmit/get-node path deterministic.

peter941221:
Small follow-up: this is still the same focused readmit-selection race fix for #1714. If you want the equality-boundary tweak split from the locking change, I can do that as a minimal follow-up.

Dosik13:
Sorry for the delay, I had some other work - I will investigate it furthermore and will make a review

Dosik13 (approved):
Thank you for the good work and sorry for the delay! Good work!

PR #1723, opened 2026-05-18
refactor: nft allowance delete

### LINKED ISSUE #1722: `DeleteTokenNftAllowanceAllSerials` should not allow approve-for-all when chained with `ApproveTokenNftAllowance` on the same `(token, spender)` pair
## Summary

`AccountAllowanceApproveTransaction._DeleteTokenNftAllowanceAllSerials` contains a logic error that inverts the meaning of the revoke operation whenever an NFT allowance entry for the same `(token, spender)` pair already exists on the transaction. Instead of emitting a wire-format `approved_for_all = false` (revoke), the SDK emits `approved_for_all = true` (grant approve-for-all).

## Affected versions

All current releases of `hiero-sdk-go` that contain the merge branch in `_DeleteTokenNftAllowanceAllSerials`.

## Affected code

`sdk/account_allowance_approve_transaction.go`, line 221 (inside `_DeleteTokenNftAllowanceAllSerials`, lines 216–235):

```go
func (tx *AccountAllowanceApproveTransaction) _DeleteTokenNftAllowanceAllSerials(
    tokenID TokenID, ownerAccountID *AccountID, spenderAccount AccountID,
) *AccountAllowanceApproveTransaction {
    for _, t := range tx.nftAllowances {
        if t.TokenID.String() == tokenID.String() {
            if t.SpenderAccountID.String() == spenderAccount.String() {
                t.SerialNumbers = []int64{}
                t.AllSerials = true   // BUG: should be false
                return tx
            }
        }
    }

    tx.nftAllowances = append(tx.nftAllowances, &TokenNftAllowance{
        TokenID:          &tokenID,
        SpenderAccountID: &spenderAccount,
        SerialNumbers:    []int64{},
        AllSerials:       false,   // correct on the "append fresh" path
        OwnerAccountID:   ownerAccountID,
    })
    return tx
}
```

Note the asymmetry: the "append a fresh entry" path correctly sets `AllSerials: false`. Only the "merge with existing entry" branch is wrong.

## Root cause

The function is a near copy-paste of `_ApproveTokenNftAllowanceAllSerials` (immediately above in the same file), where `t.AllSerials = true` is *correct* because that helper IS the approve-for-all path. When the helper was duplicated for the delete path, the boolean on the merge branch was not flipped. The "append fresh" branch below was flipped correctly, which is why the bug only manifests on the merge path.

## Impact

When a user constructs a single `AccountAllowanceApproveTransaction` and chains:

```go
NewAccountAllowanceApproveTransaction().
    ApproveTokenNftAllowance(nftID, owner, spender).             // grant a single serial
    DeleteTokenNftAllowanceAllSerials(tokenID, owner, spender).  // intent: revoke approve-for-all
    Execute(client)
```

…the SDK keeps one in-memory `TokenNftAllowance` entry per `(token, spender)` pair. The first call appends an entry with `SerialNumbers = [N]` and `AllSerials = false`. The second call enters the merge branch, sets `SerialNumbers = []`, and sets `AllSerials = true`. The serialized transaction body therefore contains a single `NftAllowance{ approved_for_all: true }` message — byte-identical to a legitimate `ApproveTokenNftAllowanceAllSerials` payload — and consensus stores it as a real approve-for-all grant.

## Fix

One-line change in `sdk/account_allowance_approve_transaction.go`:

```diff
 func (tx *AccountAllowanceApproveTransaction) _DeleteTokenNftAllowanceAllSerials(
     tokenID TokenID, ownerAccountID *AccountID, spenderAccount AccountID,
 ) *AccountAllowanceApproveTransaction {
     for _, t := range tx.nftAllowances {
         if t.TokenID.String() == tokenID.String() {
             if t.SpenderAccountID.String() == spenderAccount.String() {
                 t.SerialNumbers = []int64{}
-                t.AllSerials = true
+                t.AllSerials = false
                 return tx
             }
         }
     }
     ...
```

## Tests to add

1. **Unit test** asserting `_DeleteTokenNftAllowanceAllSerials` sets `AllSerials = false` on the merge path:
   - Construct an `AccountAllowanceApproveTransaction`, call `ApproveTokenNftAllowance(NftID{token, 3}, owner, spender)`.
   - Then call `DeleteTokenNftAllowanceAllSerials(token, owner, spender)`.
   - Assert the resulting `tx.nftAllowances` has length 1, `SerialNumbers == []`, and `AllSerials == false`.
2. **Protobuf round-trip test** confirming the serialized `NftAllowance` for the chained scenario has `approved_for_all: false`.
3. **e2e regression test** — the PoC above, adapted into the existing integration suite.

The existing unit tests for this method cover only the "no prior entry, append fresh" path, which is why this bug went uncaught.

Dosik13:
**Description**:

Simplify `AccountAllowanceApproveTransaction.DeleteTokenNftAllowanceAllSerials` so that each call produces its own `NftAllowance` entry. Repeated or chained calls on the same `(token, spender)` pair are no longer folded into a single entry.

  * Update `_DeleteTokenNftAllowanceAllSerials` to append a new `NftAllowance` entry per call instead of folding into an existing entry
  * Add unit test covering `ApproveTokenNftAllowance` followed by `DeleteTokenNftAllowanceAllSerials` on the same `(token, spender)` pair
  * Add unit test covering repeated `DeleteTokenNftAllowanceAllSerials` calls on the same `(token, spender)` pair
  * Add e2e test covering `ApproveTokenNftAllowance` followed by `DeleteTokenNftAllowanceAllSerials` on the same `(token, spender)` pair

**Related issue(s)**:

Fixes #1722

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/sdk/account_allowance_approve_transaction.go
+++ b/sdk/account_allowance_approve_transaction.go
@@ -214,16 +214,6 @@ func (tx *AccountAllowanceApproveTransaction) _ApproveTokenNftAllowanceAllSerial
 }
 
 func (tx *AccountAllowanceApproveTransaction) _DeleteTokenNftAllowanceAllSerials(tokenID TokenID, ownerAccountID *AccountID, spenderAccount AccountID) *AccountAllowanceApproveTransaction {
-	for _, t := range tx.nftAllowances {
-		if t.TokenID.String() == tokenID.String() {
-			if t.SpenderAccountID.String() == spenderAccount.String() {
-				t.SerialNumbers = []int64{}
-				t.AllSerials = true
-				return tx
-			}
-		}
-	}
-
 	tx.nftAllowances = append(tx.nftAllowances, &TokenNftAllowance{
 		TokenID:          &tokenID,
 		SpenderAccountID: &spenderAccount,
--- a/sdk/account_allowance_approve_transaction_unit_test.go
+++ b/sdk/account_allowance_approve_transaction_unit_test.go
@@ -301,3 +301,71 @@ func TestUnitDeleteTokenNftAllowanceAllSerials(t *testing.T) {
 		})
 	}
 }
+
+func TestUnitDeleteTokenNftAllowanceAllSerialsPreservesPriorSerialApproval(t *testing.T) {
+	t.Parallel()
+
+	transactionID := TransactionIDGenerate(AccountID{Account: 324})
+
+	transaction, err := NewAccountAllowanceApproveTransaction().
+		SetTransactionID(transactionID).
+		SetNodeAccountIDs(nodeAccountID).
+		ApproveTokenNftAllowance(nftID1, owner, spenderAccountID1).
+		DeleteTokenNftAllowanceAllSerials(tokenID2, owner, spenderAccountID1).
+		Freeze()
+	require.NoError(t, err)
+
+	data := transaction.build()
+
+	switch d := data.Data.(type) {
+	case *services.TransactionBody_CryptoApproveAllowance:
+		require.Equal(t, d.CryptoApproveAllowance.NftAllowances, []*services.NftAllowance{
+			{
+				TokenId:           tokenID2._ToProtobuf(),
+				Owner:             owner._ToProtobuf(),
+				Spender:           spenderAccountID1._ToProtobuf(),
+				SerialNumbers:     []int64{serialNumber1},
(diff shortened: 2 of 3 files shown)


gsstoykov (commented):
Add e2e test.

gsstoykov (approved):

PR #1732, opened 2026-05-24
fix: Re-enable fee_estimate_query example and remove mirror-sync sleeps

### LINKED ISSUE #1696: Re-enable fee_estimate_query example in run-examples and remove mirror-sync sleeps
### Problem

The `fee_estimate_query` example is currently skipped in Taskfile.yml's `run-examples` task because it flakes against a fresh local mirror node — the mirror hasn't ingested the operator's state yet when `run-examples` invokes the example back-to-back with all the others.                         
The integration tests in `sdk/fee_estimate_query_e2e_test.go` work `time.Sleep` before every `FeeEstimateQuery`. That wait is a band-aid — it shouldn't be required if the SDK or mirror node properly handles the not-yet-ingested case.                                                                                                          

### Solution

When the problem is fixed in solo - remove the `fee_estimate_query` `if` block from `Taskfile.yml`'s `run-examples` task.
**Remove `waitForMirrorNodeSync()` calls** in `sdk/fee_estimate_query_e2e_test.go` (and the `mirrorSyncDelay` constant + helper if no other callers).

### Alternatives

_No response_

Mounil2005:
**Description**:

Re-enable fee_estimate_query example in run-examples and remove mirror-sync sleeps

The `fee_estimate_query` example was previously skipped in the `run-examples` task due to intermittent failures on fresh local mirror node deployments. The underlying issue was that the mirror node hadn't ingested the operator's state when the example ran back-to-back with others.

This PR removes the band-aid `time.Sleep()` workarounds and re-enables the example by relying on the proper `waitForFeeEstimationServiceReady()` readiness probe, which polls the FeeEstimationService until it's actually ready to handle queries.

* Remove fee_estimate_query skip block from Taskfile.yml run-examples task
* Remove all waitForMirrorNodeSync() calls from fee_estimate_query_e2e_test.go (12 instances)
* Remove mirrorSyncDelay constant and waitForMirrorNodeSync() helper function
* Rely on waitForFeeEstimationServiceReady() readiness probe instead of arbitrary sleeps

**Related issue(s)**:

Fixes #1696

**Notes for reviewer**:

The `waitForFeeEstimationServiceReady()` function was already in place and working correctly. It probes the FeeEstimationService with a fresh INTRINSIC estimate request every 5 seconds until it succeeds or 10 minutes elapse. This is a proper solution compared to the previous blind 2-second sleep, as it:
- Doesn't waste time if the service is ready quickly
- Properly detects when the FeeEstimationService can handle queries
- Works across all parallel tests with a shared readiness check

**Checklist**

- [x] Documented (Issue references and comment block explain the behavior)
- [x] Tested (Integration tests now run without artificial sleeps and use proper readiness probing)

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -17,11 +17,6 @@ tasks:
                       echo "Skipping $example"
                       continue
                   fi
-                  # fee_estimate_query is flaky on a fresh local mirror node.
-                  if [ "$dir_name" == "fee_estimate_query" ]; then
-                      echo "Skipping $example (mirror node sync flake)"
-                      continue
-                  fi
                   if [ -d "$example" ]; then
 
                       pushd "$example" > /dev/null
--- a/examples/fee_estimate_query/main.go
+++ b/examples/fee_estimate_query/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"fmt"
 	"os"
+	"time"
 
 	hiero "github.com/hiero-ledger/hiero-sdk-go/v2/sdk"
 )
@@ -31,6 +32,14 @@ func main() {
 	client.SetOperator(operatorAccountID, operatorKey)
 
 	fmt.Println("FeeEstimateQuery Example (HIP-1261)")
+
+	// On a fresh solo deployment the mirror node's FeeEstimationService races
+	// the importer ingesting the genesis fee schedule, so the first few calls
+	// return HTTP 400 "Unknown transaction type". Poll until it answers or 10
+	// minutes have elapsed.
+	if err := waitForFeeEstimationService(client); err != nil {
+		panic(fmt.Sprintf("%v : fee estimation service never became ready", err))
+	}
 
 	// Step 1: Create and freeze a transfer transaction. The query auto-freezes
 	// if the transaction is not already frozen, but freezing up front lets us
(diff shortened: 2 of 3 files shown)


gsstoykov (commented):
I don't see where `waitForFeeEstimationServiceReady` is relied upon.

edit: found it mb

gsstoykov (commented):
Have you validated probing this a couple of times. Let's say 10-100?

Mounil2005:
> Have you validated probing this a couple of times. Let's say 10-100?

Yes waitForFeeEstimationServiceReady uses sync.Once so the probe runs exactly once per test suite regardless of how many parallel tests call it. It actively polls a known-good intrinsic estimate every 5 seconds for up to 10 minutes, only unblocking tests once an actual success response comes back from the mirror node. This is fundamentally different from the old fixed time.Sleep(2s) the tests won't start until the service is genuinely ready, so repeated runs should not flake.

Mounil2005:
Pls let me know in case if anything needs to be changed!

Mounil2005:
@Dosik13 @gsstoykov 
Can you pls review this and let me know in case of any changes.

Dosik13 (approved):
lgtm! thank you for the work and the patience!

PR #1756, opened 2026-06-29
fix(sdk): correct panic message in TokenSupplyType.String() and add unit test

AruneshDwivedi:
TokenSupplyType.String() panic message says TokenType.String() — copy-paste from token_type.go. Corrected to reference TokenSupplyType.

Table-driven test asserts the string mapping for both INFINITE and FINITE values.

### DIFF
--- a/sdk/token_supply_type_unit_test.go
+++ b/sdk/token_supply_type_unit_test.go
@@ -1,0 +1,27 @@
+//go:build all || unit
+
+package hiero
+
+// SPDX-License-Identifier: Apache-2.0
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestUnitTokenSupplyTypeString(t *testing.T) {
+	t.Parallel()
+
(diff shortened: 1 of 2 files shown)


sdk/token_supply_type_unit_test.go
```diff
@@ -0,0 +1,27 @@
...
+		expected   string
+	}{
+		{TokenSupplyTypeInfinite, "TOKEN_SUPPLY_TYPE_INFINITE"},
+		{TokenSupplyTypeFinite, "TOKEN_SUPPLY_TYPE_FINITE"},
+	}
+
+	for _, tt := range tests {
+		assert.Equal(t, tt.expected, tt.supplyType.String())
+	}
+}
```
Dosik13, line 27: Could you add a panic-path test? PanicsWithValue asserts the exact message, so it covers the branch actually changed and would've failed on the old TokenType.String() text.

Dosik13:
Due to inactivity I am closing the PR.

PR #1764, opened 2026-07-07
fix: keep unchanged nodes and connections on address-book refresh

### LINKED ISSUE #1762: Address-book refresh closes all consensus-node connections even when unchanged, failing in-flight requests with non-retryable `Canceled`
### Description

A long-running service submitting `EthereumTransaction`s had a submission fail instantly with:

```
retry 0/10: rpc error: code = Canceled desc = grpc: the client connection is closing
```

The transaction never left the process. The cause is client-internal: every address-book refresh (`_UpdateAddressBook` — scheduled every 24 h by default, run once at client creation, and re-triggered after any `INVALID_NODE_ACCOUNT` precheck) discards and recreates **every** consensus-node object and closes the gRPC `ClientConn` of every node whose channel has been opened, **even when the fetched address book is completely unchanged**. A request racing that close in the window between obtaining the node's channel and the gRPC connection pick fails with `ErrClientConnClosing` (`codes.Canceled`), which `_ExecutableDefaultRetryHandler` does not retry — so the whole execute loop aborts on attempt 0 of 10.

The root cause is a key-type mismatch in `_SetNetwork`'s "keep unchanged nodes" bookkeeping, present in both its loops:

1. **Removal loop** — `_GetNodesToRemove` ([`sdk/managed_network.go` L276–L286](https://github.com/hiero-ledger/hiero-sdk-go/blob/v2.80.0/sdk/managed_network.go#L276-L286)) probes the incoming node map with `v._GetKey()`, which for `_Node` is the **account-ID string** (e.g. `"0.0.3"`, [`sdk/node.go` L46–L48](https://github.com/hiero-ledger/hiero-sdk-go/blob/v2.80.0/sdk/node.go#L46-L48)). But that map is keyed by **URL** ([`sdk/network.go` L23–L35](https://github.com/hiero-ledger/hiero-sdk-go/blob/v2.80.0/sdk/network.go#L23-L35), `newNetwork[url] = node`). An account-ID string can never equal a URL key, so **every existing node is treated as "removed" and closed, every refresh**:

   ```go
   func _GetNodesToRemove(network map[string]_IManagedNode, nodes []_IManagedNode) []int {
   	nodeIndices := []int{}
   	for i, v := range slices.Backward(nodes) {
   		if _, ok := network[v._GetKey()]; !ok {   // "0.0.3" looked up among URL keys — never matches
   			nodeIndices = append(nodeIndices, i)
   		}
   	}
   	return nodeIndices
   }
   ```

2. **Add/skip loop** — the same mismatch in the other direction ([`sdk/managed_network.go` L72–L86](https://github.com/hiero-ledger/hiero-sdk-go/blob/v2.80.0/sdk/managed_network.go#L72-L86)): `newNodeKeys` is populated with account IDs but probed with the URL map key, so the skip never matches and every incoming node is appended. Today this is masked by bug 1 (everything was just removed, so appending everything is coincidentally correct), but the two must be fixed **together** — fixing only the removal side would keep the old nodes *and* append a duplicate node set on every refresh, growing `mn.nodes`/`healthyNodes` without bound.

   ```go
   	for _, node := range newNodes {
   		newNodeKeys[node._GetKey()] = true        // accountID, e.g. "0.0.3"
   		newNodeValues[node._GetAddress()] = true
   	}

   	for key, value := range network {             // key is the URL
   		_, keyOk := newNodeKeys[key]              // URL looked up in an accountID set — never matches
   		_, valueOk := newNodeValues[value._GetAddress()]
   		if keyOk && valueOk {
   			continue
   		}
   		newNodes = append(newNodes, value)
   	}
   ```

The same `_SetNetwork` works correctly for **mirror** networks because `_MirrorNode._GetKey()` returns the address — the key/URL equivalence the code assumes holds only there, which is likely why this went unnoticed.

Two aggravating factors turn the churn into a request failure:

- **In-flight requests race the close.** Per-node `channelMutex` serializes `_GetChannel` against `_Close`, but the protection ends when `_GetChannel` returns: the caller uses the returned channel for the gRPC invoke outside any lock ([`sdk/executable.go` L262](https://github.com/hiero-ledger/hiero-sdk-go/blob/v2.80.0/sdk/executable.go#L262)). A close landing between channel fetch and the connection pick yields `ErrClientConnClosing` — grpc-go returns this only **pre-send** (picker/`newClientStream`), so the request provably never reached any node. (Separately, `_SetNetwork` swaps `mn.nodes`/`mn.network`/`mn.healthyNodes` without holding `healthyNodesMutex`, which other accessors use — a data race on those fields, though not the failure reported here.)
- **`codes.Canceled` is not retryable.** [`sdk/executable.go` L389–L406](https://github.com/hiero-ledger/hiero-sdk-go/blob/v2.80.0/sdk/executable.go#L389-L406) retries only `ResourceExhausted`/`Unavailable`/`DeadlineExceeded` (plus `Internal` RST-stream), so this client-internal failure escapes to the caller immediately even though a retry would lazily re-dial via `_GetChannel` and succeed.

**What I expected to happen:**

- A refresh that fetches an address book identical to the current network should be a no-op: existing node objects and their open connections should be kept.
- A genuinely changed address book should only close the connections of nodes that were actually removed or re-addressed — ideally without failing requests already in flight.
- A pre-send, client-internal `Canceled` caused by connection recycling should be retried within the execute loop's attempt budget rather than aborting on attempt 0/10.

**Suggested fix:** use a shared `(key, address)` pair identity in **both** loops, e.g.:

```go
func _NodePairKey(n _IManagedNode) string {
	return n._GetKey() + "|" + n._GetAddress()
}

func _GetNodesToRemove(network map[string]_IManagedNode, nodes []_IManagedNode) []int {
	newPairs := make(map[string]bool, len(network))
	for _, n := range network {
		newPairs[_NodePairKey(n)] = true
	}
	nodeIndices := []int{}
	for i, v := range slices.Backward(nodes) {
		if !newPairs[_NodePairKey(v)] {
			nodeIndices = append(nodeIndices, i)
		}
	}
	return nodeIndices
}
```

and in `_SetNetwork`, replace the `newNodeKeys`/`newNodeValues` sets with the same pair identity for the add/skip check:

```go
	kept := make(map[string]bool, len(newNodes))
	for _, node := range newNodes {
		kept[_NodePairKey(node)] = true
	}

	for _, value := range network {
		if kept[_NodePairKey(value)] {
			continue
		}
		newNodes = append(newNodes, value)
	}
```

Hardening beyond that (both optional but worthwhile): treat grpc-go's pre-send `ErrClientConnClosing` as retryable in `_ExecutableDefaultRetryHandler`, and/or defer `node._Close()` for genuinely-removed nodes until in-flight requests drain.

### Steps to reproduce

The vulnerable window is small (between `_GetChannel` returning and the gRPC connection pick), so in production this surfaces as a rare, uptime-dependent failure — roughly a once-per-24 h-of-uptime lottery via the scheduled refresh. Tightening the refresh loop makes it reproducible:

```go
client := hiero.ClientForTestnet()
client.SetOperator(operatorID, operatorKey)

// Simulate the scheduled refresh with an UNCHANGED address book.
addressBook, _ := hiero.NewAddressBookQuery().
	SetFileID(hiero.FileIDForAddressBook()).
	Execute(client)

go func() {
	for {
		client.SetNetworkFromAddressBook(addressBook) // same book every time
		time.Sleep(100 * time.Millisecond)
	}
}()

for i := 0; i < 1000; i++ {
	_, err := hiero.NewAccountBalanceQuery().
		SetAccountID(operatorID).
		Execute(client)
	if err != nil {
		fmt.Println(err) // sporadically: retry 0/10: rpc error: code = Canceled
		                 // desc = grpc: the client connection is closing
	}
}
```

That every refresh recreates all nodes (the bug itself, independent of the race) can also be observed directly: call `SetNetworkFromAddressBook` twice with the same book and note that all node objects are closed and replaced both times.

### Additional context

_No response_

### Hedera network

testnet

### Version

v2.80.0

### Operating system

None

Dosik13:
**Description**:

Fix the address-book refresh needlessly closing and recreating every consensus node's gRPC connection, even when the fetched address book is unchanged. The churn let in-flight requests race a connection close and fail pre-send with a non-retryable `Canceled` ("the client connection is closing").

* Match existing nodes against the incoming address book by address and confirm the key, so an unchanged refresh keeps the current nodes
* Skip re-adding nodes already present so their open connections are preserved and only removed or re-addressed nodes are closed
* Add unit tests covering node and connection reuse across an unchanged refresh, using real consensus nodes since the managed-network mocks set key == address

**Related issue(s)**:

Fixes #1762

**Notes for reviewer**:

**Checklist**

- [X] Documented (Code comments, README, etc.)

### DIFF
--- a/sdk/managed_network.go
+++ b/sdk/managed_network.go
@@ -75,10 +75,7 @@ func (mn *_ManagedNetwork) _SetNetwork(network map[string]_IManagedNode) error {
 	}
 
 	for key, value := range network {
-		_, keyOk := newNodeKeys[key]
-		_, valueOk := newNodeValues[value._GetAddress()]
-
-		if keyOk && valueOk {
+		if newNodeKeys[value._GetKey()] && newNodeValues[key] {
 			continue
 		}
 
@@ -284,7 +281,8 @@ func _GetNodesToRemove(network map[string]_IManagedNode, nodes []_IManagedNode) 
 	nodeIndices := []int{}
 
 	for i, v := range slices.Backward(nodes) {
-		if _, ok := network[v._GetKey()]; !ok {
+		incoming, ok := network[v._GetAddress()]
+		if !ok || incoming._GetKey() != v._GetKey() {
 			nodeIndices = append(nodeIndices, i)
 		}
 	}
--- a/sdk/network_unit_test.go
+++ b/sdk/network_unit_test.go
@@ -153,3 +153,85 @@ func TestUnitConcurrentNodeGetChannel(t *testing.T) {
 	network._ReadmitNodes()
 	require.Equal(t, len(nodes), len(network.healthyNodes))
 }
+
+// Re-applying an unchanged network must keep the existing nodes and their open
+// connections rather than closing and recreating them.
+
+func TestUnitNetworkRefreshRemovesNoNodesWhenUnchanged(t *testing.T) {
+	t.Parallel()
+
+	network := _NewNetwork()
+	err := network.SetNetwork(newNetworkMockNodes())
+	require.NoError(t, err)
+
+	// Incoming address book identical to the installed nodes, keyed by address
+	// the same way _Network.SetNetwork builds it.
+	incoming := make(map[string]_IManagedNode, len(network.nodes))
+	for _, node := range network.nodes {
+		incoming[node._GetAddress()] = node
+	}
+
+	require.Empty(t, _GetNodesToRemove(incoming, network.nodes),
+		"an unchanged address book must not mark any node for removal")
+}
+
+func TestUnitNetworkRefreshPreservesNodeObjects(t *testing.T) {
+	t.Parallel()
+
+	nodes := newNetworkMockNodes()
+	network := _NewNetwork()
+	err := network.SetNetwork(nodes)
+	require.NoError(t, err)
+
+	before := nodesByAccount(&network)
+
+	// Re-apply the identical address book, as the scheduled network update does.
+	err = network.SetNetwork(nodes)
+	require.NoError(t, err)
+
+	after := nodesByAccount(&network)
+
+	require.Equal(t, len(before), len(after))
+	for account, node := range before {
+		require.Samef(t, node, after[account],
+			"node 0.0.%d was recreated by an unchanged refresh", account)
+	}
+}
+
+func TestUnitNetworkRefreshPreservesNodeConnections(t *testing.T) {
+	t.Parallel()
+
+	nodes := newNetworkMockNodes()
+	network := _NewNetwork()
+	err := network.SetNetwork(nodes)
+	require.NoError(t, err)
+
+	// Open a node's channel, standing in for an in-flight request holding a live
+	// connection. grpc.NewClient is lazy, so nothing is dialed.
+	account := AccountID{0, 0, 3, nil, nil, nil}
+	node, ok := network._GetNodeForAccountID(account)
+	require.True(t, ok)
+	_, err = node._GetChannel(NewLogger("", LoggerLevelError))
+	require.NoError(t, err)
+	require.NotNil(t, node.channel)
+
+	err = network.SetNetwork(nodes)
+	require.NoError(t, err)
+
+	refreshed, ok := network._GetNodeForAccountID(account)
+	require.True(t, ok)
+	require.Same(t, node, refreshed,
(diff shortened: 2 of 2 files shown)


gsstoykov (approved):
LGTM

PR #1790, opened 2026-07-24
fix: honor payment tx id if present

### LINKED ISSUE #1782: [Intermediate]: Honor SetPaymentTransactionID in query payment generation and re-enable ContractInfoQuery test
## 🧩 Intermediate Friendly

This issue is a good fit for contributors who are familiar with the
Hiero Go SDK integration testing setup and have a working understanding
of Solo-based local environments.

Intermediate Issues often involve:

- Extending existing test coverage (unit and/or e2e in `sdk/`)
- Understanding how integration tests interact with `NewIntegrationTestEnv` and network configuration
- Making focused improvements without major architectural changes

This task fixes a query-execution regression where a user-supplied payment transaction ID is silently ignored, and re-enables the unit test that was skipped because of it.

---

## 🐞 Problem Description

`TestUnitContractInfoQueryGetTransactionIDMock` in `sdk/contract_info_query_unit_test.go:115` is skipped:

```go
// sdk/contract_info_query_unit_test.go:116
t.Skip("Skipping test as it is currently broken with the addition of generating new payment transactions for queries")
```

The test sets an explicit payment transaction ID via `SetPaymentTransactionID(transactionID)` and asserts that the payment transaction attached to the query carries that same ID. It fails because query payment generation **ignores** the user-set ID and always generates a fresh one:

```go
// sdk/query.go:261-278
func (q *Query) generatePayments(client *Client, cost Hbar) (*services.Transaction, error) {
    var tx *services.Transaction
    var err error
    nodeID := q.nodeAccountIDs._GetCurrent()
    txnID := TransactionIDGenerate(client.operator.accountID)   // <-- always regenerates, ignores SetPaymentTransactionID
    tx, err = _QueryMakePaymentTransaction(
        txnID,
        nodeID.(AccountID),
        client.operator,
        cost,
    )
    ...
}
```

Meanwhile the public setter and getter exist and store the value:

```go
// sdk/query.go:177
func (q *Query) SetPaymentTransactionID(transactionID TransactionID) *Query {
    q.paymentTransactionIDs._Clear()._Push(transactionID)._SetLocked(true)
    ...
}

// sdk/query.go:163
func (q *Query) GetPaymentTransactionID() TransactionID { ... }
```

So `SetPaymentTransactionID` is effectively a no-op during execution — the setter is dead. This is a genuine behavior bug, not just a broken test. Note `generatePayments` is shared by **all** query types, so the fix is centralized in `query.go`.

---

## 💡 Expected Outcome

- `generatePayments` uses the caller-supplied payment transaction ID when one has been set via `SetPaymentTransactionID`, and only falls back to `TransactionIDGenerate(...)` when none is set.
- `SetPaymentTransactionID` becomes functional again across all queries.
- `TestUnitContractInfoQueryGetTransactionIDMock` has its `t.Skip(...)` removed and passes.
- No regressions for queries that don't set a payment transaction ID (the default auto-generate path is preserved).

---

## 🧠 Implementation Notes

Suggested approach:

- In `generatePayments` (`sdk/query.go:261`), derive `txnID` from the stored ID when present:
  - The IDs live in `q.paymentTransactionIDs` (a `_LockableSlice`); `GetPaymentTransactionID()` (`sdk/query.go:163`) already returns the current one, guarded by `!q.paymentTransactionIDs._IsEmpty()`.
  - If a payment transaction ID has been set, use it; otherwise keep `TransactionIDGenerate(client.operator.accountID)`.
- Re-read the test at `sdk/contract_info_query_unit_test.go:115` to confirm it passes once the ID is honored — it unmarshals the payment `BodyBytes` and compares the transaction ID against the one it set.
- **Regression risk:** because `generatePayments` is shared, verify the full unit suite still passes — particularly other `*_query_unit_test.go` mock tests and any that call `GetCost` (which also triggers payment generation).
- Consider whether the ID should advance per attempt/node (the slice is lockable/advanceable). Keep behavior consistent with how other executables reuse vs. regenerate IDs on retry; if unsure, note the question in the PR for maintainer input rather than guessing.
- Do **not** modify generated code under `services/`.

> **Alternative (if maintainers prefer the current always-generate behavior):** instead of changing `generatePayments`, rewrite `TestUnitContractInfoQueryGetTransactionIDMock` to assert the new behavior (a payment tx is attached with a generated ID) and remove the skip. Flag this trade-off in the PR — the recommended path above treats the dead setter as the bug to fix.

---

## ✅ Acceptance Criteria

- [ ] `generatePayments` honors a payment transaction ID set via `SetPaymentTransactionID`, falling back to generation only when none is set
- [ ] The `t.Skip(...)` at `sdk/contract_info_query_unit_test.go:116` is removed
- [ ] Test passes:
  - `go test ./sdk -tags="unit" -v -run TestUnitContractInfoQueryGetTransactionIDMock -timeout 9999s`
- [ ] No regressions across the full unit suite (`generatePayments` is shared by all queries):
  - `go test ./sdk -tags="unit" -v -timeout 9999s`
- [ ] No test is left disabled/skipped as a workaround
- [ ] `go build ./...` succeeds
- [ ] `golangci-lint run` passes
- [ ] `go fmt ./...` produces no changes
- [ ] Follow existing project conventions (see `.claude/CLAUDE.md`)
- [ ] DCO sign-off on every commit (`git commit -s -S`)
- [ ] Pass all CI checks

---

## 📚 Additional Context

- Skipped test: `sdk/contract_info_query_unit_test.go:115`
- Payment generation: `sdk/query.go:261` (`generatePayments`), setter `sdk/query.go:177`, getter `sdk/query.go:163`
- `_QueryMakePaymentTransaction` is where the ID is baked into the payment transaction body

If you have questions, the community is happy to help:
https://discord.com/channels/905194001349627914/1337424630289404015

radkomih:
**Description:**

Honor SetPaymentTransactionID when generating query payments so an explicitly-set payment transaction ID is used instead of being silently replaced by a freshly generated one.

- Update Query.generatePayments to use the payment transaction ID set via SetPaymentTransactionID when present, falling back to auto-generation otherwise
- Re-enable the previously skipped TestUnitContractInfoQueryGetTransactionIDMock

**Related issue(s):**

Fixes #1782

**Notes for reviewer:**

The setter stored the ID in q.paymentTransactionIDs, but generatePayments always called TransactionIDGenerate(...) and never read it back, so the setter was a no-op across all query types. The fix reads the current ID from the slice when it is non-empty.

**Checklist**

### DIFF
--- a/sdk/contract_info_query_unit_test.go
+++ b/sdk/contract_info_query_unit_test.go
@@ -113,7 +113,6 @@ func TestUnitContractInfoQueryMock(t *testing.T) {
 }
 
 func TestUnitContractInfoQueryGetTransactionIDMock(t *testing.T) {
-	t.Skip("Skipping test as it is currently broken with the addition of generating new payment transactions for queries")
 	t.Parallel()
 
 	transactionID := TransactionIDGenerate(AccountID{Account: 123})
--- a/sdk/query.go
+++ b/sdk/query.go
@@ -262,7 +262,14 @@ func (q *Query) generatePayments(client *Client, cost Hbar) (*services.Transacti
 	var tx *services.Transaction
 	var err error
 	nodeID := q.nodeAccountIDs._GetCurrent()
-	txnID := TransactionIDGenerate(client.operator.accountID)
+
+	var txnID TransactionID
+	if !q.paymentTransactionIDs._IsEmpty() {
+		txnID = q.paymentTransactionIDs._GetCurrent().(TransactionID)
+	} else {
+		txnID = TransactionIDGenerate(client.operator.accountID)
+	}
+
 	tx, err = _QueryMakePaymentTransaction(
 		txnID,
 		nodeID.(AccountID),


Dosik13 (approved):
lgtm!

PR #1798, opened 2026-08-05
fix: cancel per-attempt gRPC contexts promptly

### LINKED ISSUE #1797: Per-attempt gRPC contexts are not canceled promptly in _Execute
### Description

`_Execute` creates a deadline context for every gRPC attempt inside its retry loop:

```go
ctx, cancel := context.WithTimeout(context.Background(), grpcDeadline)
defer cancel()
```

Because the defer is registered in the scope of the entire _Execute function, the cancellation function for a completed attempt is not called immediately after its RPC returns.

When an RPC completes before its deadline and execution subsequently retries, the completed attempt's timeout context may remain alive until either:

- its deadline expires; or
- _Execute returns and executes the deferred cancellation functions.
- 
The default retry count and gRPC deadline keep the impact bounded, so this is primarily a resource-lifecycle issue rather than a functional correctness issue. However, it becomes more noticeable when a relatively long custom gRPC deadline is used.

Expected behavior: each attempt's context should be canceled immediately after its RPC completes.

### Steps to reproduce

1. Check out the current `main` branch.
2. Open `sdk/executable.go`.
3. Configure an execution with:
   - at least two attempts;
   - a relatively long gRPC deadline, such as one minute;
   - a mocked first RPC response that causes `_Execute` to retry, such as `BUSY`;
   - a second RPC response that succeeds.
4. Capture the context passed to the first mocked RPC.
5. During the second attempt, before `_Execute` returns, inspect the first context:

   ```go
   firstAttemptContext.Err()
   ```

6. Observe that it returns nil, showing that the completed first attempt has not been explicitly canceled.
7. After _Execute returns, observe that the context reports context.Canceled.

This also follows directly from the current control flow: defer cancel() is declared inside the retry loop but belongs to the surrounding _Execute function scope, so it does not run at the end of an individual iteration.

Expected result:
```go
firstAttemptContext.Err() == context.Canceled
```

immediately after the first RPC completes.
Actual result:

```go
firstAttemptContext.Err() == nil
```

until its deadline expires or _Execute returns.

### Additional context

The default configuration uses a 10-second gRPC deadline and up to 10 attempts, so the impact is bounded under default settings. A custom longer deadline can keep resources associated with completed attempts alive for longer.

A possible fix is to move the individual RPC invocation into a small helper or closure:

```go
func executeRequest(...) (...) {
    ctx, cancel := context.WithTimeout(context.Background(), grpcDeadline)
    defer cancel()

    // Execute one query or transaction RPC.
}
```

This keeps defer cancel() for safe cleanup while limiting its scope to one RPC attempt. The retry and response-handling behavior does not need to change.

### Hedera network

other

### Version

`main (868a1011, based on v2.82.0)`

### Operating system

macOS

zjuzhongwen:
**Description**:

`_Execute` previously called `defer cancel()` directly inside its retry loop. Since the defer belonged to the outer function scope, cancellation for completed attempts was delayed until `_Execute` returned.

This could keep timeout-related resources alive longer than necessary when an RPC completed before its configured deadline.

The RPC invocation is now scoped to a small closure, allowing `defer cancel()` to run at the end of each attempt.

**Related issue(s)**: 

Fixes #1797 

**Notes for reviewer**:

**Checklist**


sdk/executable.go
```diff
@@ -380,6 +364,29 @@ func _Execute(client *Client, e Executable) (interface{}, error) {
 	return &services.Response{}, errPersistent
 }
 
+func _ExecuteRequest(method _Method, grpcDeadline time.Duration, protoRequest interface{}) (interface{}, []byte, error) {
```
Dosik13, line 368: The extraction itself looks good and scoping `defer cancel()` to one attempt is exactly right. Can we add a short doc comment - something like this:

```suggestion
// _ExecuteRequest executes a single gRPC attempt within its own timeout context.
func _ExecuteRequest(method _Method, grpcDeadline time.Duration, protoRequest interface{}) (interface{}, []byte, error) {
```

sdk/executable.go
```diff
@@ -380,6 +364,29 @@ func _Execute(client *Client, e Executable) (interface{}, error) {
...
+		return resp, marshaledResponse, nil
+	}
+
+	resp, err := method.transaction(ctx, protoRequest.(*services.Transaction))
+	if err != nil {
+		return resp, nil, err
+	}
+
+	marshaledResponse, _ := protobuf.Marshal(resp)
+	return resp, marshaledResponse, nil
```
Dosik13, line 384: you can make this cleaner by as both response types implement `proto.Message` and the query/transaction if branches are identical so you can remove the duplication (something like this):

```go
	var resp protobuf.Message
	var err error
	if method.query != nil {
		resp, err = method.query(ctx, protoRequest.(*services.Query))
	} else {
		resp, err = method.transaction(ctx, protoRequest.(*services.Transaction))
	}
	if err != nil {
		return resp, nil, err
	}

	marshaledResponse, _ := protobuf.Marshal(resp)
	return resp, marshaledResponse, nil
```

sdk/executable_unit_test.go
```diff
@@ -0,0 +1,31 @@
...
+	"context"
+	"testing"
+	"time"
+
+	"github.com/hiero-ledger/hiero-sdk-go/v2/proto/services"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc"
+)
+
+func TestUnitExecuteRequestCancelsContext(t *testing.T) {
```
Dosik13: This test asserts the context is canceled after `_ExecuteRequest` returns, which mostly re-tests that `defer` works - it can only fail if the `defer cancel()` line itself is deleted. The actual bug is a retry-loop property - attempt 1's context should already be canceled while attempt 2 is still running.

Mock-server tests for `_Execute` already live in `mock_test.go`, so I'd suggest writing this there instead of adding a new file. `TestUnitMockQuery` already responds `BUSY` first and then `OK` - you can follow that pattern, capture the first attempt's context in the handler, and assert it's canceled by the time the second attempt arrives. That's the repro from the issue itself, it goes through the real retry loop, and it also covers the `transaction` branch of `_ExecuteRequest`, which currently has no test (only `query` is exercised here).

If you keep this direct test as well, a `require.NotNil(t, requestContext)` before the last assertion would make the failure readable if the stub were ever not invoked (right now a nil context would panic).

Dosik13 (changes requested):
Thank you for the contribution — the fix is correct, and extracting the attempt into its own function is a clean way to scope the `defer cancel()`. Summing up my comments:

1. Add a one-line doc comment on the helper so the per-attempt scope doesn't get inlined away later.
2. Deduplicate the two branches so the marshal + return appears once.
3. Please move the test into the existing `mock_test.go` and drive it through `_Execute` (`BUSY` → `OK`) — that checks the real retry behaviour and also covers the `transaction` path.

One more small thing: please change the issue reference to `Fixes #1797` so it auto-closes. Happy to approve after that!

sdk/executable.go
zjuzhongwen: Done

zjuzhongwen:
> Thank you for the contribution — the fix is correct, and extracting the attempt into its own function is a clean way to scope the `defer cancel()`. Summing up my comments:
> 
> 1. Add a one-line doc comment on the helper so the per-attempt scope doesn't get inlined away later.
> 2. Deduplicate the two branches so the marshal + return appears once.
> 3. Please move the test into the existing `mock_test.go` and drive it through `_Execute` (`BUSY` → `OK`) — that checks the real retry behaviour and also covers the `transaction` path.
> 
> One more small thing: please change the issue reference to `Fixes #1797` so it auto-closes. Happy to approve after that!

@Dosik13  Thanks for the review! 

I’ve addressed all the comments:

- Added a doc comment explaining the per-attempt scope of _ExecuteRequest.
- Deduplicated the query/transaction branches so marshaling and returning happen once.
- Moved the regression test into mock_test.go and now drive it through _Execute with a transaction retry (BUSY → OK). 

The test verifies that the first attempt’s context is already canceled when the second attempt begins.

Thanks again!

Dosik13 (approved):
Thank you very much!

PR #1822, opened 2026-08-31
fix: reject entity ID strings with an empty checksum

chiliec:
## What

`_AccountIDFromString`, `_ContractIDFromString` and `_IdFromString` extract the optional checksum by splitting the input on `-`, but never check that the checksum part is non-empty. So an ID with a dangling separator and no checksum — e.g. `"0.0.123-"` — is silently accepted as a valid ID that carries an **empty-string** checksum.

That's more than cosmetic. The empty checksum is stored as a non-nil `*string`, which then bypasses the `checksum == nil` guard in `ValidateChecksum`:

```go
if id.checksum == nil {
    return errChecksumMissing
}
if tempChecksum.correctChecksum != *id.checksum { // "esxsf" != "" -> misleading error
```

So a missing checksum is reported as a *wrong* checksum (`given checksum: ` empty) instead of the intended `errChecksumMissing`.

## Fix

After the `-` split, reject an empty checksum with the same error already used for a malformed `{shard}.{realm}.{num}-{checksum}` format, in all three parsers.

## Testing

Added `sdk/entity_id_empty_checksum_unit_test.go`:
- `"0.0.123-"`, `"0.0.0-"`, `"1.2.3-"` are now rejected by `AccountIDFromString` / `ContractIDFromString` / `TokenIDFromString`
- a well-formed `"0.0.123-esxsf"` still parses and preserves the `esxsf` checksum

```
$ go test ./sdk/ -tags unit -run 'AccountID|ContractID|TokenID|FileID|TopicID|ScheduleID|Checksum'
ok  	github.com/hiero-ledger/hiero-sdk-go/v2/sdk
```

`gofmt` and `go vet -tags unit` clean. First-time contributor here — happy to adjust.


sdk/id.go
```diff
@@ -33,6 +33,10 @@ func _AccountIDFromString(s string) (shard int, realm int, num int, checksum *st
 			return 0, 0, 0, nil, nil, nil, fmt.Errorf("expected {shard}.{realm}.{num}-{checksum}")
 		}
 
+		if values[1] == "" {
+			return 0, 0, 0, nil, nil, nil, fmt.Errorf("expected {shard}.{realm}.{num}-{checksum}")
+		}
```
Dosik13:
```suggestion
    if len(values) > 2 || values[1] == "" {
        return 0, 0, 0, nil, nil, nil, fmt.Errorf("expected {shard}.{realm}.{num}-{checksum}")
    }
```

sdk/id.go
```diff
@@ -98,6 +102,10 @@ func _ContractIDFromString(s string) (shard int, realm int, num int, checksum *s
 			return 0, 0, 0, nil, nil, fmt.Errorf("expected {shard}.{realm}.{num}-{checksum}")
 		}
 
+		if values[1] == "" {
+			return 0, 0, 0, nil, nil, fmt.Errorf("expected {shard}.{realm}.{num}-{checksum}")
+		}
```
Dosik13:
```suggestion
		if  len(values) > 2  || values[1] == "" {
			return 0, 0, 0, nil, nil, fmt.Errorf("expected {shard}.{realm}.{num}-{checksum}")
		}
```

sdk/id.go
```diff
@@ -166,6 +174,10 @@ func _IdFromString(s string) (shard int, realm int, num int, checksum *string, e
 			return 0, 0, 0, nil, fmt.Errorf("expected {shard}.{realm}.{num}-{checksum}")
 		}
 
+		if values[1] == "" {
+			return 0, 0, 0, nil, fmt.Errorf("expected {shard}.{realm}.{num}-{checksum}")
+		}
+
```
Dosik13: The same as above

sdk/entity_id_empty_checksum_unit_test.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// A trailing "-" with an empty checksum (e.g. "0.0.123-") must be rejected, not
+// silently parsed as an ID carrying an empty-string checksum. An empty checksum
+// pointer bypasses the `checksum == nil` guard in ValidateChecksum and turns a
+// "missing checksum" case into a misleading "wrong checksum" comparison.
```
Dosik13: Trim to a line or two — the test name already says what it does, and the repo keeps
comments sparse. The mechanism explanation is better placed in the commit message.

sdk/entity_id_empty_checksum_unit_test.go
```diff
@@ -0,0 +1,36 @@
+//go:build all || unit
```
Dosik13: `sdk/account_id_unit_test.go` already owns `TestUnitAccountIDChecksumFromString` and
`TestUnitChecksum`; this fits naturally there rather than in a new single-test file so you can move the test there

Dosik13 (changes requested):
Thanks for the PR, left some suggestions

sdk/id.go
chiliec: Done — merged into the `len(values) > 2` check in 427bd9c.

sdk/id.go
chiliec: Done in 427bd9c.

sdk/id.go
chiliec: Done in 427bd9c.

sdk/entity_id_empty_checksum_unit_test.go
chiliec: Trimmed to a single line and moved the mechanism detail to the commit message (427bd9c).

sdk/entity_id_empty_checksum_unit_test.go
chiliec: Moved into `sdk/account_id_unit_test.go` and deleted the standalone file (427bd9c).

chiliec:
Thanks for the review @Dosik13! All addressed in 427bd9c:

- Merged the empty-checksum guard into the existing `len(values) > 2` check in all three parsers (`_AccountIDFromString`, `_ContractIDFromString`, `_IdFromString`).
- Trimmed the test comment to one line and moved the mechanism detail to the commit message.
- Relocated the test into `sdk/account_id_unit_test.go` next to `TestUnitChecksum` / `TestUnitAccountIDChecksumFromString`, and removed the standalone file.

`gofmt` clean and the affected unit tests pass.

Dosik13:
@chiliec, everything looks good but you need to sign your commit for the DCO check to pass

chiliec:
Thanks @Dosik13 — added the missing `Signed-off-by` trailer to the review-fix commit (the second commit had dropped it). Force-pushed and DCO is green now.

Dosik13 (approved):
Thank you very much!

